PR #154, opened 2015-02-22
Include song rating in MPRIS metadata

### LINKED ISSUE #153: Include loved/banned song status in MPRIS track metadata
My panel (Taffybar) currently shows Pithos's currently playing song via the panel's MPRIS widget, and I'd like to tweak that widget to also show the loved-or-not status of the current song, but Pithos doesn't expose that in the song metadata. I'd like to add support for that, but I'm not sure if I should use the existing `xesam:userRating` field (http://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index22h4) with some mapping (e.g., 0.0 -> banned, 0.5 -> normal, 1.0 -> loved) or introduce a new `pithos:rating` field with a string value. What do you think?

bcat:
MPRIS metadata for the currently playing track now includes a custom
`pithos:rating` property with a string rating status: empty string (no
rating), `love`, or `ban`. In practice, `ban` will not be seen since
Pithos skips the current song after banning it, but this rating will
appear if Pithos exposes the full playlist via the
`org.mpris.MediaPlayer2.TrackList` API.

Pithos also skips a song immediately after marking it tired, so
including a `pithos:tired` property didn't seem to make sense, but that
might also be useful if MPRIS track lists are implemented in the future.

Closes #153.

### DIFF
--- a/pithos/mpris.py
+++ b/pithos/mpris.py
@@ -38,6 +38,7 @@ class PithosMprisService(dbus.service.Object):
         self.song_changed()
         
         self.window.connect("song-changed", self.songchange_handler)
+        self.window.connect("song-rating-changed", self.ratingchange_handler)
         self.window.connect("play-state-changed", self.playstate_handler)
         
     def playstate_handler(self, window, state):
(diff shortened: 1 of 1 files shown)


TingPing:
Thanks!

PR #158, opened 2015-03-13
Added error checking for older versions of Gstreamer

jamiesonbecker:
Added error checking for older versions of Gstreamer

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -22,13 +22,24 @@ import logging, argparse
 import signal
 
 import gi
-gi.require_version('Gst', '1.0')
 from gi.repository import Gst, GstPbutils, GObject, Gtk, Gdk, Pango, GdkPixbuf, Gio, GLib
 import contextlib
 import html
 import math
 import urllib.request, urllib.error, urllib.parse
 import json
+
+try:
+    gi.require_version('Gst', '1.0')
+except ValueError, e:
+    if e[0].startswith("Namespace Gst not available for version"):
+        print "Received error %s from Gstreamer." % e[0]
+        print "This workstation may not have the version of gstreamer or GTK required."
+        print "See also https://github.com/pithos/pithos/issues/32"
+        sys.exit(1)
+    else:
+        raise
+
 if sys.platform != 'win32':
     from dbus.mainloop.glib import DBusGMainLoop
     DBusGMainLoop(set_as_default=True)


pithos/pithos.py
```diff
@@ -22,13 +22,24 @@
...
 import html
 import math
 import urllib.request, urllib.error, urllib.parse
 import json
+
+try:
+    gi.require_version('Gst', '1.0')
+except ValueError, e:
+    if e[0].startswith("Namespace Gst not available for version"):
+        print "Received error %s from Gstreamer." % e[0]
```
TingPing: Based on #159 you already know this is invalid.

jamiesonbecker: Actually I didn't realize at all that this was related. The default Python on my system that I used for python setup.py build etc is python 2.7. Your build seems to have zero checks for versions or missing packages, so I do consider it broken, at least to some extent, certainly not very inviting to potential contributors. :)

TingPing: Except it does use python3 by default and you had to call it with python2 manually... https://github.com/pithos/pithos/blob/master/setup.py#L1

Also setuptools is absolutely useless for checking for deps and since they are runtime deps it would be counter productive for packagers to force deps anyway.

jamiesonbecker: I git cloned and typed python setup.py build. This of course ignores the 
hashbang line. The app itself should check for missing libs at runtime, 
since the system you build with is often not the same system it's 
deployed on. However, not trying to start an argument - I'm just 
naturally inclusive of even relatively old versions of *, and we may 
just have different philosophies of what's valuable or worth putting in 
extra time for. No worries :)

TingPing: >The app itself should check for missing libs at runtime

Well uh, yea it does, you get that for free. Plugins handle it nicely and pithos refuses to start basically, as expected.

jamiesonbecker: Originally I was using the Pithos available for Wheezy but it wouldn't 
continue playing on some songs and wasn't sure if it was related to 
fresh purchase of Pandora One, so decided to build from source. 
Following the natural path of cloning and python setup.py build, I built 
and installed it without any issues (using default python on setup.py). 
Tried to launch newly built pithos (/usr/local/bin) and received gi 
version traceback (see also Issue 32). After adding a nicer version 
check (above) for gstreamer, another import failure on html; pip 
installed that. (A list of external python libraries/deps might be 
useful on the wiki.) Finally, the abort on urllib imports only in Python 
3 clued me in that it wasn't a Python2/3 app and realized further time 
investment was not useful. App looks good and useful but apparently not 
for me :)

PR #163, opened 2015-03-25
Sort stations by name

andrewgaul:
This allows users to quickly find a single station when they have many
configured.  This affects both the main window station picker and the
manage stations dialog.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -437,11 +437,12 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.current_station = None
         selected = None
 
-        for i in self.pandora.stations:
+        def extract_name(station): return station.name
+        for i in sorted(self.pandora.stations, key=extract_name):
             if i.isQuickMix and i.isCreator:
                 self.stations_model.append((i, "QuickMix"))
         self.stations_model.append((None, 'sep'))
-        for i in self.pandora.stations:
+        for i in sorted(self.pandora.stations, key=extract_name):
             if not (i.isQuickMix and i.isCreator):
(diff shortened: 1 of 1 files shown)


gregsheremeta:
Thanks! Looks fine to me. Been wanting this myself :)

@TingPing ?

TingPing:
My plan was to eventually make the station list a nice popover with a sort button so users could chose what they wanted and sort. Hardcoding always alphabetical is not idea imo.

gregsheremeta:
@TingPing would you be interested if @andrewgaul made it configurable with an option?

TingPing:
> @TingPing would you be interested if @andrewgaul made it configurable with an option?

As a short term solution maybe. The second problem is the station list already has sort functionality. This sort of breaks that window having the ability to show the sorted by date added (currently unsorted).

andrewgaul:
I don't advocate this approach as the best solution, merely an available improvement.  No one has offered a better implementation since I submitted #13 almost two years ago.  I do not understand the use case of sorting by other criteria like date added and most music managers like Rhythmbox at least default to alphabetical sort.

TingPing:
So instead of fixing the things I brought up a year ago you just reopened the same thing with absolutely no improvements?

I just now created a branch [wip/popover](https://github.com/pithos/pithos/tree/wip/popover) for what I think is the proper solution to this. It's not complete but doesn't hardcode sorting things or add extra random options.

gregsheremeta:
Thanks for the PR -- we're going to implement this with GTK popovers. Hopefully soon :)

PR #166, opened 2015-03-30
Better UI loop handling

JasonLG1979:
As it sits Pithos starts a new UI loop every time play is pressed(maybe even more often?) and never destroys old loops. They stack up over time creating a small memory leak of sorts. This patch makes Pithos handle the UI loop more intelligently by starting/creating it when needed and stopping/destroying it when it's not needed (like when Pithos is paused). Credit for the code goes to the Exaile devs as it's pretty much a copy and paste of their code with a little tweaking to make it work with Pithos.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -245,7 +245,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.gstreamer_error = ''
         self.waiting_for_playlist = False
         self.start_new_playlist = False
-
+        self.ui_loop_timer_id = None
         self.worker = GObjectWorker()
         self.art_worker = GObjectWorker()
 
@@ -508,8 +508,8 @@ class PithosWindow(Gtk.ApplicationWindow):
     def play(self):
         if not self.playing:
             self.playing = True
+            self.create_ui_loop()
         self.player.set_state(Gst.State.PLAYING)
-        GLib.timeout_add_seconds(1, self.update_song_row)
         self.playpause_image.set_from_icon_name('media-playback-pause-symbolic', Gtk.IconSize.SMALL_TOOLBAR)
         self.update_song_row()
         self.emit('play-state-changed', True)
@@ -520,6 +520,7 @@ class PithosWindow(Gtk.ApplicationWindow):
 
     def pause(self):
         self.playing = False
+        self.destroy_ui_loop()
         self.player.set_state(Gst.State.PAUSED)
         self.playpause_image.set_from_icon_name('media-playback-start-symbolic', Gtk.IconSize.SMALL_TOOLBAR)
         self.update_song_row()
@@ -534,6 +535,7 @@ class PithosWindow(Gtk.ApplicationWindow):
             self.emit("song-ended", prev)
 
         self.playing = False
+        self.destroy_ui_loop()
         self.player.set_state(Gst.State.NULL)
         self.emit('play-state-changed', False)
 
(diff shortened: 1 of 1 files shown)


pithos/pithos.py
```diff
@@ -843,7 +845,18 @@ def update_song_row(self, song = None):
         if song:
             self.songs_model[song.index][1] = self.song_text(song)
             self.songs_model[song.index][2] = self.song_icon(song) or ""
-        return self.playing
+        if self.playing:
```
TingPing: No reason for this change.

pithos/pithos.py
```diff
@@ -843,7 +845,18 @@ def update_song_row(self, song = None):
...
+            return True
+
+    def create_ui_loop(self):
+        if self.ui_loop_timer_id is not None:
```
TingPing: This likely should not compare against None. If it succeeds it will always return > 0 and if it fails it returns 0, in which case you *want* a new timeout.

pithos/pithos.py
```diff
@@ -520,6 +520,7 @@ def user_pause(self, *ignore):
 
     def pause(self):
         self.playing = False
+        self.destroy_ui_loop()
```
TingPing: This call is unnecessary. Once update_song_row returns False (aka paused) it will already destroy it.

pithos/pithos.py
```diff
@@ -843,7 +845,18 @@ def update_song_row(self, song = None):
...
+        if self.playing:
+            return True
+
+    def create_ui_loop(self):
+        if self.ui_loop_timer_id is not None:
+            return
+        self.ui_loop_timer_id = GLib.timeout_add(1000, self.update_song_row)
+        
+    def destroy_ui_loop(self):
+        if self.ui_loop_timer_id is not None:
```
TingPing: Same note here about None.

pithos/pithos.py
```diff
@@ -534,6 +535,7 @@ def stop(self):
             self.emit("song-ended", prev)
 
         self.playing = False
+        self.destroy_ui_loop()
```
TingPing: Same note here about it being auto destroyed.

TingPing:
So I think this commit can be simplified down to just adding `self.ui_loop_timer_id` and then just checking for that in `play()` since removal should be fine, unless I missed something.

pithos/pithos.py
JasonLG1979: The loop dies 1 sec after it return False. Destroying the loop kills it right then. So I guess it really doesn't matter.

pithos/pithos.py
JasonLG1979: I just used what the Exaile devs used. They used None. I haven't tried to modify that part of the code as it has NEVER failed.

JasonLG1979: *Failed to work

JasonLG1979: self.ui_loop_timer_id has to be reset somehow(to None or 0 or whatever) otherwise when the loop dies it won't be able to be restarted since the ID is not cleared.

Could be as easy as - 

if not self.playing:
    self.ui_loop_timer_id = None (or 0 or whatever)
return return self.playing

in update_song_row

pithos/pithos.py
JasonLG1979: If the loop is not manually destroyed if you click play more than once in 1 sec a new loop is created  before the old loop dies resulting in multiple loops. The whole point of the patch is to prevent that.

JasonLG1979:
This code works perfectly well as is. I've been running with it for months with no issues. No need to change it.

TingPing:
I did make those None changes. But the automatic destruction was not sufficient so this was correct. Your commit message was slightly off but I left it.

JasonLG1979:
Commit message was slightly off?

TingPing:
It did destroy old loops but only when paused for long enough.

JasonLG1979:
It didn't destroy them it just let them die,lol

PR #184, opened 2015-05-10
Disable download buffering

JasonLG1979:
Disable download buffering to avoid deadlocks in Gstreamer.


gregsheremeta:
Looks good to me, meth hamster.

TingPing:
Is that seriously all needed to "fix" the problems so many have had?

gregsheremeta:
Not the buffering issues, but it does fix the deadlock issue afaict.

JasonLG1979:
The buffering issue is most likely caused from starting/putting Gstreamer in a play state before the pipeline is prerolled. Starting Gstreamer in a paused state and then waiting for the async-done message before setting it to play should fix the buffering issue. I'll do a pull request for that next.

PR #187, opened 2015-05-10
Buffer stop bug fix.

JasonLG1979:
1. Start pipeline in paused state.
2. Only use on_gst_buffering to start a timeout(1/10 sec) to actively query the buffer percent, and once the async-done message is received act on the buffer query accordingly.
3. Got rid of the buffer % in the buffering message as it was pretty much useless since it's only there for a fraction of a sec for mp3 streams and goes from 0 straight to 100 for aac streams.
4. Added a few switching values to make it all possible.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -107,7 +107,7 @@ class PlayerStatus (object):
   def reset(self):
     self.async_done = False
     self.began_buffering = None
-    self.buffer_percent = 100
+    self.buffer_percent = 0
     self.pending_duration_query = False
 
 
@@ -177,6 +177,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.stations_model = Gtk.ListStore(GObject.TYPE_PYOBJECT, str)
 
         Gst.init(None)
+        self._query_buffer = Gst.Query.new_buffering(Gst.Format.PERCENT)
         self._query_duration = Gst.Query.new_duration(Gst.Format.TIME)
         self._query_position = Gst.Query.new_position(Gst.Format.TIME)
         self.player = Gst.ElementFactory.make("playbin", "player");
@@ -195,7 +196,8 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.player_status = PlayerStatus()
 
         self.stations_dlg = None
-
+        self.buffering = False
+        self.player_paused = False
         self.playing = False
         self.current_song_index = None
         self.current_station = None
@@ -210,6 +212,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.waiting_for_playlist = False
         self.start_new_playlist = False
         self.ui_loop_timer_id = 0
+        self.buffer_watch_loop_timer_id = 0
         self.worker = GObjectWorker()
         self.art_worker = GObjectWorker()
 
@@ -451,7 +454,9 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.player_status.reset()
 
         self.player.set_property("uri", self.current_song.audioUrl)
-        self.play()
+        self.player_status.began_buffering = time.time()
+        self.player.set_state(Gst.State.PAUSED)
+        self.buffering = True
         self.playcount += 1
 
         self.current_song.start_time = time.time()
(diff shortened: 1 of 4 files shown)


pithos/pithos.py
```diff
@@ -630,6 +640,14 @@ def query_duration(self):
         _, duration = self._query_duration.parse_duration()
         return duration
 
+    def query_buffer(self):
+        buf_stat = self.player.query(self._query_buffer)
+        if buf_stat:
+            _, percent = self._query_buffer.parse_buffering_percent()
```
TingPing: Minor style thing: `percent = self._query_buffer.parse_buffering_percent()[1]`

pithos/pithos.py
```diff
@@ -195,7 +196,8 @@ def init_core(self):
         self.player_status = PlayerStatus()
 
         self.stations_dlg = None
-
+        self.buffering = False
+        self.player_paused = False
```
TingPing: Maybe make use of `player_status` or similar instead of adding new variables?

JasonLG1979: I make use of the player_status thing a lot in my fork. I was just going for the bare minimum as far as changes. Baby steps and all for commits right,lol.

pithos/pithos.py
```diff
@@ -721,28 +739,23 @@ def on_gst_tag(self, bus, message):
...
-            else:
-                self.player.set_state(Gst.State.PLAYING)
-                logging.debug("Took %.3f to buffer", time.time() - self.player_status.began_buffering)
-                self.player_status.began_buffering = None
-        self.player_status.buffer_percent = percent
-        self.update_song_row()
-        logging.debug("Buffering (%i%%)", self.player_status.buffer_percent)
+        self.create_buffer_watch_loop()
+
+    def react_to_buffer(self):
```
TingPing: Maybe something like `on_buffer_timer` makes more sense as a name?

pithos/pithos.py
JasonLG1979: Whatever's clever. I really don't have much as far as style.

TingPing:
If you pause during buffering it will just play anyway.

JasonLG1979:
That's how stock Pithos works too.

pithos/pithos.py
```diff
@@ -721,28 +739,23 @@ def on_gst_tag(self, bus, message):
...
+                self.player.set_state(Gst.State.PAUSED)
+                self.buffering = True
+            else:   
+                self.play()
+                self.buffering = False
+                self.destroy_buffer_watch_loop()
+                self.create_ui_loop()
+                logging.debug("It took %.3f seconds to buffer.", time.time() - self.player_status.began_buffering)
+            self.update_song_row()
+            logging.debug("Buffering (%i%%)", buffer_percent)
```
TingPing: Move this debug message up so they come out in the logical order.

```
DEBUG - pithos:react_to_buffer:757 - Buffering (0%)
DEBUG - pithos:react_to_buffer:755 - It took 1.145 seconds to buffer.
DEBUG - pithos:react_to_buffer:757 - Buffering (100%)
```

TingPing:
> That's how stock Pithos works too.

That is not true. You introduced it on line 751 where you always play ignoring `self.playing` as it used to. (or now your new variable `self.player_paused`

JasonLG1979:
Oops. I'll fix that and all the other stuff tomorrow.

TingPing:
On a related note to that it now takes two presses to pause to actually pause while buffering. I didn't track down the exact cause to that but I really think this would be cleaner if you didn't add two variables for playing state. `self.playing` could have a third state `None` when initialized, I presume that is the entire reason you created `self.player_paused`.

JasonLG1979:
I have a more complete implementation of the whole state awareness deal in my other branch. It works flawlessly. I'll clean up this branch a little tomorrow

JasonLG1979:
I was basically working backwards from my other branch. A couple things fell through the cracks I guess. Oops again.

gregsheremeta:
Used the "start pipeline in paused state" idea in https://github.com/pithos/pithos/releases/tag/1.1.1

That indeed fixed the buffering weirdness people were seeing. Thanks!

PR #188, opened 2015-05-10
fix "set_header_func() takes exactly 3 arguments (2 given)" on GTK < …

gregsheremeta:
…3.12

Works great on Fedora 20 and 21.


gregsheremeta:
@TingPing is this the proper way to work around stuff like this in python / gtk?

TingPing:
Seems good to me, the issue is actually in pygobject though if you want to update your comments.

gregsheremeta:
updated comments. thanks.

PR #189, opened 2015-05-10
Fix sending song ratings to last.fm.

JasonLG1979:
Now the sending of ratings (Love or Ban) is done when the rating happens via the rating-changed signal. Otherwise if the user did not actively rate the song during playback the rating will be reported at the end of the song.

### DIFF
--- a/pithos/plugins/lastfm.py
+++ b/pithos/plugins/lastfm.py
@@ -50,6 +50,7 @@ class LastfmPlugin(PithosPlugin):
         self.pylast = pylast
         self.worker = get_worker()
         self.is_really_enabled = False
+        self.user_rated = False
         self.preferences_dialog = LastFmAuth(self.pylast, self.window.preferences, "lastfm_key", self.window)
         self.preferences_dialog.connect('delete-event', self.auth_closed)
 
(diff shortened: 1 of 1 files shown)


TingPing:
This was removed on purpose many years ago. The issue was that the meanings of love/ban on pandora do not directly match the rating system on last.fm. Now that we have a system in place for plugins having preferences I would not be against adding this as an option though.

JasonLG1979:
They both have Love and Ban. What does Ban mean in Last.fm?

JasonLG1979:
Love and Ban seem to pretty much mean the same thing in Pithos/Pandora and Last.fm.
http://www.last.fm/forum/34905/_/369421

TingPing:
This decision was well before my time, I'll open up the launchpad issues later to find the discussion.

JasonLG1979:
A ui option in the last.fm plugin seems to be the best idea I guess?

TingPing:
https://bugs.launchpad.net/pithos/+bug/636600

tl;dr is it spams your last.fm account with "Favorited tracks" when all the user wanted to do was a pandora station to play more similar songs.

> A ui option in the last.fm plugin seems to be the best idea I guess?

An option should be fine.

JasonLG1979:
Well I guess out this on hold until We/I can put a ui option in there and figure out how to check if a song has already been loved in the past (as not to spam)

JasonLG1979:
*hold up on this

PR #195, opened 2015-05-18
Don't scrobble ads.

JasonLG1979:
Pithos assumes any song less than 45 sec is an ad and changes the description in the UI to reflect that. But the bogus info is still accessible. We don't want to scrobble bogus songs.

### DIFF
--- a/pithos/plugins/lastfm.py
+++ b/pithos/plugins/lastfm.py
@@ -102,7 +102,7 @@ class LastfmPlugin(PithosPlugin):
     def scrobble(self, song):
         duration = song.get_duration_sec()
         position = song.get_position_sec()
-        if duration > 30 and (position > 240 or position > duration/2):
+        if duration > 45 and (position > 240 or position > duration/2):
             logging.info("Scrobbling song")
             mode = self.pylast.SCROBBLE_MODE_PLAYED
             source = self.pylast.SCROBBLE_SOURCE_PERSONALIZED_BROADCAST


TingPing:
Makes more sense to check `song.is_ad` than some magic variable you hope is kept in sync.

JasonLG1979:
Whatever's clever... Magic variable? all I did was change the minimum song duration that can be scrobbled from 30 sec to 45 secs...

TingPing:
Magic variable as in without context "45" or "30" is meaningless, `song.is_ad` is obvious.

JasonLG1979:
It's pretty obvious "if duration > 45" means if the duration is less than 45... But I didn't write the function. I'm guessing that function predates the ad check. I was going for the least possible change to achieve the desired result.

gregsheremeta:
s/Magic variable/Magic number/

JasonLG1979:
> s/Magic variable/Magic number/

I don't know what that means?

gregsheremeta:
It means 'replace'. Hardcoding a number in a program is called a 'magic number,' not a 'magic variable.'

`$string =~ s/regex/replacement/g;`

http://www.regular-expressions.info/perl.html

JasonLG1979:
> It means 'replace'. Hardcoding a number in a program is called a 'magic number,' not a 'magic variable.'

Oh, Ok.

JasonLG1979:
The original functions follows last.fm scrobbling spec.
http://www.last.fm/api/scrobbling

"A track should only be scrobbled when the following conditions have been met:

The track must be longer than 30 seconds.
And the track has been played for at least half its duration, or for 4 minutes (whichever occurs earlier.)"

I just changed 30 to 45 because Pithos assumes "songs" less than 45 sec are ads.

TingPing:
> The original functions follows last.fm scrobbling spec.
> http://www.last.fm/api/scrobbling

So leave the 30 sec check, but check if we consider it an ad.

JasonLG1979:
1. What you want adds code for no reason with no added benefit. We already need to fetch the duration anyway to determine if at least half the song has played.
2. If you throw a "if not song.is_ad" in there the 30 is pointless since if it's not an ad is has to be at longer than 45 secs.

TingPing:
>What you want adds code for no reason with no added benefit.

You seem to come up with this argument everytime I critique your code... When I comment about relatively minor things like syntax I do so because I want the code to be easy to maintain, clear to read, handle edge cases, etc. Yes as is both ways absolutely have the same result and that is not the point at all.

>If you throw a "if not song.is_ad" in there the 30 is pointless since if it's not an ad is has to be at longer than 45 secs.

That is based on the idea that our ad checking methods never change. I am not convinced hard-coding a specific length is the best way to do that and it could easily change tomorrow.

JasonLG1979:
It's hardcoded now and has been for god knows how long. You complain when I send to much in a merge request. You complain when I change 1 number in a function.

TingPing:
I am not complaining, I appreciate your contributions and would like to keep the project of always improving quality.

TingPing:
@JasonLG1979 Just one last follow up, all I am doing is a "code review", an objective look at the code, it has nothing to do with you personally  and I would point out equally minor details on @gregsheremeta's code. The goal is to improve code, not to hurt feelings or complain...

JasonLG1979:
If you would have explained that sooner I would have been like "OK"

TingPing:
@JasonLG1979 Yea sorry. Most of code I review tends to be from developers familiar with others picking away at their work and tone doesn't really come across very well.

JasonLG1979:
I'm just a wee babe in code land. Be gentle, lol!!!

PR #197, opened 2015-05-18
Removing unnecessary stuff

JasonLG1979:
We can set self.playing to None in stop. No need to do it in start_song


TingPing:
:+1:

PR #200, opened 2015-05-19
Get song duration, bitrate and codec from Pandora

JasonLG1979:
Get song duration, bitrate and codec from Pandora for use in the UI. It makes for a smoother UI experience IMHO as that info is instantly available as apposed to waiting for Pithos/Gstreamer to parse the data.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -626,7 +626,9 @@ class PithosWindow(Gtk.ApplicationWindow):
       pos_stat = self.player.query(self._query_position)
       if pos_stat:
         _, position = self._query_position.parse_position()
-        return position
+      else:
+          position = 0
+      return position
 
     def query_duration(self):
       dur_stat = self.player.query(self._query_duration)
@@ -639,14 +641,14 @@ class PithosWindow(Gtk.ApplicationWindow):
       self.player_status.async_done = True
       if self.player_status.pending_duration_query:
         self.current_song.duration = self.query_duration()
-        self.current_song.duration_message = self.format_time(self.current_song.duration)
+        self.current_song.ad_duration_message = self.format_time(self.current_song.duration)
         self.check_if_song_is_ad()
         self.player_status.pending_duration_query = False
 
     def on_gst_duration_changed(self, bus, message):
       if self.player_status.async_done:
         self.current_song.duration = self.query_duration()
-        self.current_song.duration_message = self.format_time(self.current_song.duration)
+        self.current_song.ad_duration_message = self.format_time(self.current_song.duration)
         self.check_if_song_is_ad()
       else:
         self.player_status.pending_duration_query = True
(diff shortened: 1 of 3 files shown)


JasonLG1979:
Combined with @TingPing's buffer spinner the UI will be like butter.

TingPing:
You can get rid of `gst_tag_handler` then. Also you still make use of the duration info gst has, is that intended to mix and match the two?

pithos/pithos.py
```diff
@@ -639,14 +641,14 @@ def on_gst_async_done(self, bus, message):
...
-        self.current_song.duration_message = self.format_time(self.current_song.duration)
+        self.current_song.ad_duration_message = self.format_time(self.current_song.duration)
         self.check_if_song_is_ad()
         self.player_status.pending_duration_query = False
 
     def on_gst_duration_changed(self, bus, message):
       if self.player_status.async_done:
         self.current_song.duration = self.query_duration()
-        self.current_song.duration_message = self.format_time(self.current_song.duration)
+        self.current_song.ad_duration_message = self.format_time(self.current_song.duration)
```
TingPing: Whats the purpose of ad_duration_message if its set to the same thing?

JasonLG1979:
No we really don't need the tag handler anymore but it might not be a bad idea to keep it around for debug purposes?

We still need the gstreamer duration to tell if a song is an ad. Pandora lies about duration of ads to.

They're mixed in song_text because if it's an ad we want to use the gstreamer duration if not we can use the duration provided by Pandora.

TingPing:
> We still need the gstreamer duration to tell if a song is an ad. Pandora lies about duration of ads to.

I think for now it makes more sense to just trust pandora on codec/bitrate, and trust gstreamer on duration. There is little point in mixing the two when they may not match.

JasonLG1979:
Fair enough.

JasonLG1979:
What about the tag handler? I say maybe keep it but make the debug message a little more useful/less spammy

TingPing:
I don't see much value in it but that's fine. Maybe only hook that event if log level == DEBUG to improve performance, since that is sort of the point of these changes right.

JasonLG1979:
After thinking about it I agree, the tag handler is pretty useless if we get the codec and bitrate from Pandora. It's a whole lot of code just to spam the debug messages.

I'm going to close this and start over. No duration from Pandora, just the bitrate and codec, and nix the tag handler. Sound good?

TingPing:
> I'm going to close this and start over. No duration from Pandora, just the bitrate and codec, and nix the tag handler. Sound good?

Sounds good. It would be nice for you to look into `git` more deeply to avoid multiple PRs though ;)

JasonLG1979:
Will do. It was just easier to start fresh than undo everything.

PR #201, opened 2015-05-20
Show that a song has been scrobbled in the UI.

JasonLG1979:
Shows "Scrobbled to Last.fm" in the song message field after a song has been scrobbled. As it sits now the only way to know that a song has been scrobbled is running pithos -v in the terminal or going to last.fm to check.

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -330,6 +330,8 @@ class Song(object):
         self.songExplorerUrl = d['songExplorerUrl']
         self.artRadio = d['albumArtUrl']
 
+
+        self.scrobbled = False
         self.bitrate = None
         self.is_ad = None  # None = we haven't checked, otherwise True/False
         self.tired=False
@@ -444,4 +446,3 @@ class SearchResult(object):
             self.artist = d['artistName']
         elif resultType == 'artist':
             self.name = d['artistName']
-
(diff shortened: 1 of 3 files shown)


TingPing:
Not sure I like the idea of plugin specific properties inside of pandora.py. I just finally made that plugin fully optional. That said an indication of scrobble status would be nice and I'm not sure a better way for a plugin to add that.

JasonLG1979:
It was the simplest way I could figure to do it? If the plugin isn't enabled it will stay false and won't be shown.

gregsheremeta:
Good idea -- makes sense.

Agree with @TingPing. There needs to be a clear interface between pithos and it's plugins. Pithos shouldn't internally know about any plugin details.

Ideally there'd be a `set_status_message` method that's exposed by pithos, and then plugins are allowed to call that to set a message. @JasonLG1979 that make sense? Gonna close this, but I think it's easy for you to redo it :)

TingPing:
> Ideally there'd be a set_status_message method that's exposed by pithos, and then plugins are allowed to call that to set a message

I might look into that after some UI things on my todo list if @JasonLG1979 is not interested.

JasonLG1979:
I was just thinking the same thing. A generic place plugins can put messages.

gregsheremeta:
Exactly. Think of it more as an "interface" -- pithos exposing an interface so that plugins can **ask** pithos to place a message somewhere, if it so desires :)

For example, a music player's interface is well-known -- play, pause, stop, skip, etc buttons. Humans can press those buttons and usually get the action they're looking for. But whether it's an MP3 player, Pandora, cassette player, or whatever -- humans don't need to know about the internals. They just know that pressing a button will usually get them the action they're looking for.

JasonLG1979:
I'll give it a shot.

PR #202, opened 2015-05-20
update_song_row may as well always return True

JasonLG1979:
update_song_row may as well always return True since we manually create and destroy the UI update loop and since self.paying is now a 3 way switch.


TingPing:
:+1:

PR #206, opened 2015-05-22
use go-down-symbolic in station popover

JasonLG1979:
instead of view-sort-ascending-symbolic. go-down-symbolic looks nice and doesn't require GTK 3.16. As stated before kinda silly to depend on a specific version of GTK for just one icon. Especially since as of yet most popular distros don't even ship 3.16 in a stable release.

### DIFF
--- a/pithos/StationsPopover.py
+++ b/pithos/StationsPopover.py
@@ -30,7 +30,7 @@ class StationsPopover(Gtk.Popover):
         self.search = Gtk.SearchEntry()
         self.sorted = False
         sort = Gtk.ToggleButton.new()
-        sort.add(Gtk.Image.new_from_icon_name("view-sort-ascending-symbolic", Gtk.IconSize.BUTTON))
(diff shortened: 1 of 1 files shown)


TingPing:
Like I said earlier if we want to bundle the icon that is fine. I dislike changing the icon to something with a completely different meaning though.

JasonLG1979:
Meh. I guess I don't put a lot of meaning into an arrow pointing down,lol? I push the button it sorts the station list alphabetically (from top to bottom of course so go-down seems to kinda fits???) Bundle in the intended icon then I guess. I just don't like the missing icon icon. Looks ugly.

TingPing:
Well the icon names are sort of abstract right. So a *view-sort-ascending* icon in theory may represent something much different than *go-down*. Since there is no harm in bundling it I think that is an easy choice.

JasonLG1979:
Cool cool bundle a way then man. I hate that missing icon icon makes it look like something is broken(and technically it is.)

PR #207, opened 2015-05-22
Use Pandora logo if the album art url is no good.

JasonLG1979:
1st choice is of course the actual album cover. 2nd choice is a generic Pandora logo and 3rd choice is the default blue gradient.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -56,6 +56,7 @@ except ImportError:
 
 ALBUM_ART_SIZE = 96
 ALBUM_ART_X_PAD = 6
+GENERIC_LOGO = 'http://cont-sv5-1.pandora.com/images/public/devicead/1/n/r/a/daarcmdywarn1_500W_500H.jpg'
 
 class CellRendererAlbumArt(Gtk.CellRenderer):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


pithos/pithos.py
```diff
@@ -56,6 +56,7 @@
 
 ALBUM_ART_SIZE = 96
 ALBUM_ART_X_PAD = 6
+GENERIC_LOGO = 'http://cont-sv5-1.pandora.com/images/public/devicead/1/n/r/a/daarcmdywarn1_500W_500H.jpg'
```
TingPing: This looks like an auto generated URL I would not trust existing for long.

JasonLG1979:
http://cont-sv5-1.pandora.com/images/public/devicead/1/n/r/a/daarcmdywarn1_500W_500H.jpg
Still good.

It would have to be updated if it were ever to become no good.

JasonLG1979:
It's the url that's served up with the multiple stream warning.

TingPing:
That url doesn't even work for me...

JasonLG1979:
Funny works fine for me?

JasonLG1979:
screenshot from 2015-05-22 14 46 19

JasonLG1979:
screenshot from 2015-05-22 14 49 06

TingPing:
Hmm, it works in another browser but not chrome. Anyway I still feel funny about the URL and...

> It would have to be updated if it were ever to become no good.

Is not valid for any project with a release cycle. There is no point in adding it if it may not work for 80% of our users in 6mo.

JasonLG1979:
> Is not valid for any project with a release cycle. There is no point in adding it if it may not work for 80% of our users in 6mo.

Fair point. Just trying to find something better than the gradient. I'll see if I can't find something from Pandora that is legally OK to use.

JasonLG1979:
Never mind can't ship there logo in the app according to their brand guide without written consent. I was trying to side step that by just basically linking to a publicly accessible url. But never mind I guess. Would you receptive to either, 1. Using a default media icon. 2. A redesign of the "default_icon" image or 3. Using the Pithos logo maybe? If none of those thing sound like they're worth it, no big deal.

TingPing:
I am not in love with the current fallback icon so any changes you have in mind might be interesting. I don't think the pithos logo would fit in though.

pithos/pithos.py
JasonLG1979: It still exists,lol!!!

PR #209, opened 2015-05-23
Tame the UI updates during buffering

JasonLG1979:
AAC's spam the buffer messages with 0% and duplicate messages causing upwards of 1000+ buffer messages in less than 2 secs in some cases. This will calm it down a little, and make the buffer message in the UI a little more useful as it won't appear to be stuck on 0 most of the time.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -812,6 +812,8 @@ class PithosWindow(Gtk.ApplicationWindow):
             else:
                 logging.debug("Buffer recovery. User paused")
             self.player_status.began_buffering = None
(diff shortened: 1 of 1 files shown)


JasonLG1979:
@gregsheremeta and/or @TingPing this one could use some testing, but it seems to do the trick.

gregsheremeta:
@JasonLG1979 I'm bogged down in real life, but maybe I can get to these this weekend.

JasonLG1979:
Right on. Just wondering if you guys weren't finally getting tired of me ;P

Issue #183, opened 2015-05-02
Stop is only used internally in Pithos. It should not be exposed to the user.


JasonLG1979:
Stop is only used internally in Pithos. It isn't exposed to the user anywhere else. It shouldn't be exposed to the user in _mpris.py. It should either be mapped to pause or passed(my vote) or some logic needs to be come up with so that if you press stop and then play again it skips to the next song and plays it. Right now if you hit stop and then play again it starts the song over. Which IMHO is not expected behavior for a "Radio" player.

JasonLG1979:
No comments/thoughts on this? Anyone?

TingPing:
Pausing seems reasonable. Shame mpris requires exposing stop of some sort.

gregsheremeta:
Agree, I vote for mapping it to pause. Definitely should be no starting a song over :-/

JasonLG1979:
The mpris plugin needs a lot of work. While messing with adding track time to it I found that basically every time ANY property is updated it takes the nuclear route and updates EVERYTHING. Aside from actually changing the metadata every action pretty much tells mpris that there's a new song.

JasonLG1979:
> Definitely should be no starting a song over :-/

That's actually what it's suppose to do according to spec, start the song over. That just doesn't fit for Pithos.

http://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Method:Stop

PR #210, opened 2015-05-24
Map stop to pause in mpris

JasonLG1979:
And fixed a tiny typo.


JasonLG1979:
@gregsheremeta and/or @TingPing - We talked about this in https://github.com/pithos/pithos/issues/183. Again this has been sitting for a wile should I just close it?

PR #211, opened 2015-05-24
Fix edge case in scrobbler function

JasonLG1979:
1. We should check if song.is_ad == False because song.is_ad is a 3 way switch where None means we haven't checked yet or detection has failed, in either case we don't want to try to scrobble the track.

2. We shouldn't even try to check the duration or position if song.is_ad != False, song.is_ad == False means that the duration and position is available(duration is needed to check for ads currently). Otherwise in the edge case where a song is skipped before duration and position is available we'll get an error.

### DIFF
--- a/pithos/plugins/lastfm.py
+++ b/pithos/plugins/lastfm.py
@@ -100,13 +100,15 @@ class LastfmPlugin(PithosPlugin):
             logging.info("Sending song rating to last.fm")
 
     def scrobble(self, song):
-        duration = song.get_duration_sec()
-        position = song.get_position_sec()
-        if not song.is_ad and duration > 30 and (position > 240 or position > duration/2):
-            logging.info("Scrobbling song")
-            mode = self.pylast.SCROBBLE_MODE_PLAYED
-            source = self.pylast.SCROBBLE_SOURCE_PERSONALIZED_BROADCAST
-            self.worker.send(self.scrobbler.scrobble, (song.artist, song.title, int(song.start_time), source, mode, duration, song.album))
+        if song.is_ad == False:
+            duration = song.get_duration_sec()
+            position = song.get_position_sec()
+            if duration is not None and position is not None:
+                if duration > 30 and (position > 240 or position > duration/2):
+                    logging.info("Scrobbling song")
+                    mode = self.pylast.SCROBBLE_MODE_PLAYED
+                    source = self.pylast.SCROBBLE_SOURCE_PERSONALIZED_BROADCAST
+                    self.worker.send(self.scrobbler.scrobble, (song.artist, song.title, int(song.start_time), source, mode, duration, song.album))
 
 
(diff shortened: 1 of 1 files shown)


TingPing:
> where None means we haven't checked yet or detection has failed

I did that on purpose, if we fail to detect it I assumed it is a fine assumption that it's not an ad, as that is the more likely situation. Is this situation even possible atm?

> We shouldn't even try to check the duration or position if song.is_ad

That seems reasonable.

TingPing:
Well actually we should handle duration being None anyway, not necessarily associate that with is_ad.

JasonLG1979:
> Is this situation even possible atm?

I have yet to see ad detection fail since it was moved to the async-done message function. But
it's possible to skip a song fast enough that song.is_ad == None and the duration is not available yet leading to a type error because we're then trying to compare None to an int.

> Well actually we should handle duration being None anyway, not necessarily associate that with is_ad.

We should, but there's also no reason to even proceed if song.is_ad != False. Explicit is better than implicit.

TingPing:
> We should, but there's also no reason to even proceed if song.is_ad != False.

Hmm, Yea that is good.

JasonLG1979:
> Hmm, Yea that is good.

I guess we could throw a "if duration and position is not None:" in there in case in the future ad detection is not tied to duration.

JasonLG1979:
Something more like that @TingPing?

pithos/plugins/lastfm.py
```diff
@@ -100,13 +100,15 @@ def send_rating(self, song, rating):
...
-        position = song.get_position_sec()
-        if not song.is_ad and duration > 30 and (position > 240 or position > duration/2):
-            logging.info("Scrobbling song")
-            mode = self.pylast.SCROBBLE_MODE_PLAYED
-            source = self.pylast.SCROBBLE_SOURCE_PERSONALIZED_BROADCAST
-            self.worker.send(self.scrobbler.scrobble, (song.artist, song.title, int(song.start_time), source, mode, duration, song.album))
+        if song.is_ad == False:
+            duration = song.get_duration_sec()
+            position = song.get_position_sec()
+            if duration and position is not None:
```
TingPing: That might not do what you think:

```py
if duration is not None and position is not None:
```

JasonLG1979: What's the difference?

TingPing: Currently it does `if duration` *and* `if position is not None`. In this case I there is no real issue with not comparing to `None`. Just letting you know that doesn't actually work how english might expect it to.

JasonLG1979: I guess it could be a problem if duration was ever false?(it never should be) Thanks for the calcification.

JasonLG1979: Better?

Issue #213, opened 2015-05-27
User Pause not honored during song/station change or buffering.


JasonLG1979:
Expected behavior(IMHO) - 
Whenever I click pause(even during buffering) Pithos should say paused until I click play. If I click pause then skip a song or load a new station the next/new song should buffer but not start playing until I click play.

Current  behavior - 
Pause during buffering works sometimes. If I click pause then skip to the next song/load a new station the song buffers and start playing automatically even though I did not click play.

IMHO Pause should never be overridden. 

I can fix this. I just want your thoughts on this @TingPing and @gregsheremeta before I go down that rabbit hole.

TingPing:
I don't like that behavior. The act of hitting next or changing stations implies "I want to listen to something else", at least to me.

gregsheremeta:
I agree with @JasonLG1979 on it, but we may be wrong. It'd be interesting to see if there's a convention or spec on it (since ultimately, these buttons constitute a ubiquitous interface). What do media players, DVD players, VLC, Windows Media Player, etc. do?

gregsheremeta:
VLC starts playing after skip while paused.

JasonLG1979:
Ok, I guess we shouldn't care about pause going into a new song. What about being able to pause during buffering?

TingPing:
Pausing during buffering should be respected as-is.

JasonLG1979:
> Pausing during buffering should be respected as-is.

? But it's not respected.

gregsheremeta:
Yeah I didn't follow @TingPing's comment either. If I'm buffering and I hit pause, it should stay paused. If it doesn't, it's a bug.

JasonLG1979:
For aac streams I think the reason it may not be respected is that the buffer throws a couple 100% messages before it actually settles down. Causing it to go back and forth between buffering and not. I think something maybe getting lost along the way as the code in the buffering function is trying to protect user pause. 

For mp3 streams buffering usually takes less than 1 sec(for me anyway) so you pretty much have to be a ninja to actually hit pause while it's actually buffering.

JasonLG1979:
> VLC starts playing after skip while paused.

Clementine, same behavior.

TingPing:
https://github.com/pithos/pithos/commit/af94c55d9960c1a991c19cdd21846a01f6c3a5f5 went out of its way to make sure it respects pause during buffering, but as @JasonLG1979 mentioned the timing is rather specific.

JasonLG1979:
[...]

PR #214, opened 2015-05-27
Fix pause during buffering

JasonLG1979:


TingPing:
Like I've mentioned before I hate the idea of multiple variables representing play and pause, it makes no sense.

JasonLG1979:
I don't like it either but since you made playing a 3 way switch you can't rely on "if self.playing:" for the playpause button.

TingPing:
If we need more than 3 states just factor this into an enum, like:

```python
from enum import Enum
class PlayState(Enum):
    waiting=0
    paused=1
    user_paused=2
    playing=3
```

JasonLG1979:
Cool.

JasonLG1979:
Will that allow for overlapping states. Because it's possible to be waiting for a song to start and be paused at the same time or not be paused and waiting for a song to start or not be waiting for a song to start and be paused or not waiting for a song to start and not be paused. We had something like this when we has song_started to represent only if the song had started or not and playing to represent if it was paused or not.

TingPing:
>  Because it's possible to be waiting for a song to start and be paused at the same time

No its not, that is what #213 was all about. Currently 'waiting' always means waiting to play, if that is changed to pause it just never starts, and if its set to play well just play. There is no need for overlap afaict.

JasonLG1979:
Yes it is. If I hit pause during buffering I'm paused while waiting for a song to start

JasonLG1979:
The old song_started basically meant done initially buffering

JasonLG1979:
The current code in doesn't work because it conflates 2 unrelated things. If we're done initially buffering and if we are currently playing.

JasonLG1979:
@TingPing, I changed it so that self.playing represents playing or paused and self.initially_buffered represents(you guessed it) if the song has been initially buffered with those 2 variable it's possible to represent the 4 possible states of:
1. Being in a Play state(or at least wanting to be) while buffering.
2. Being in a Paused state(or again at least wanting to be) while buffering.
3. Being in a Play state while not buffering.
4. Being in a Paused state while not buffering.

PR #218, opened 2015-06-05
Cache song_text and song_icon

JasonLG1979:
Basically if the text or icon hasn't changed don't update it.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -879,8 +879,10 @@ class PithosWindow(Gtk.ApplicationWindow):
         if song is None:
             song = self.current_song
         if song:
-            self.songs_model[song.index][1] = self.song_text(song)
-            self.songs_model[song.index][2] = self.song_icon(song) or ""
+            if self.songs_model[song.index][1] != self.song_text(song):
+                self.songs_model[song.index][1] = self.song_text(song)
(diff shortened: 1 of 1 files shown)


JasonLG1979:
@TingPing We briefly talked about this on the IRC. I believe it was your idea.

pithos/pithos.py
```diff
@@ -879,8 +879,10 @@ def update_song_row(self, song = None):
         if song is None:
             song = self.current_song
         if song:
-            self.songs_model[song.index][1] = self.song_text(song)
-            self.songs_model[song.index][2] = self.song_icon(song) or ""
+            if self.songs_model[song.index][1] != self.song_text(song):
+                self.songs_model[song.index][1] = self.song_text(song)
+            if self.songs_model[song.index][2] != self.song_icon(song) or "":
```
TingPing: Remove the extra `or ""`

JasonLG1979: I guess I don't know which one is extra?

TingPing: The one in the if statement.

JasonLG1979:
Done.

JasonLG1979:
Turns out after testing it by putting a simple "print("update text")" and "print("update icon")" after each if statement the extra or "" is in fact not extra. Without the icon part doesn't work. It still updates the icon even when not needed.

PR #231, opened 2015-10-04
Miniature Mode

dspeyer:
Hello all,

This is a "mini-mode" for Pithos, in which the window shrinks to only 32 pixels high, but still contains the vital controls (including a button to return to normal mode).  This fits my usage patterns much better than the normal UI.  I tend to have Pithos playing in the background while I work on something else. With this UI, I can stick it at the top of the screen and still have essentially maximized windows.  I suspect it would also be good for embedding in toolbars, but I haven't tested that.

I've been using it for a while now and it's been working well for me, so I thought I'd send it upstream to you.

### DIFF
--- a/debug.py
+++ b/debug.py
@@ -20,7 +20,7 @@ if sys.platform != 'win32':
 
     # Enable verbose logging and test mode
     if len(sys.argv) == 1:
-        sys.argv.append('-tv')
+        sys.argv.append('-tvv')
 
 from pithos import pithos
 
--- a/pithos/data/ui/PithosWindow.ui
+++ b/pithos/data/ui/PithosWindow.ui
@@ -15,9 +15,9 @@
     <property name="default_width">500</property>
     <property name="default_height">360</property>
     <property name="icon_name">pithos</property>
+    <signal name="configure-event" handler="on_configure_event" swapped="no"/>
     <signal name="destroy" handler="on_destroy" swapped="no"/>
     <signal name="key-press-event" handler="on_kb_playpause" swapped="no"/>
-    <signal name="configure-event" handler="on_configure_event" swapped="no"/>
     <child>
       <object class="GtkBox" id="vbox1">
         <property name="visible">True</property>
@@ -37,7 +37,6 @@
                   <object class="GtkBox" id="playcontrol_box">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
-                    <property name="homogeneous">True</property>
                     <child>
                       <object class="GtkButton" id="playpause_button">
                         <property name="visible">True</property>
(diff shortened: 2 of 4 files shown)


TingPing:
Need to rebase on master.

TingPing:
It's an interesting idea I am not against but using it few minor complaints:

- No way to move window once in mini mode
- Duplicate sound title in full mode
- mini mode modifies size of all text

I have yet to actually check out any of the code though.

dspeyer:
I've hidden the redundant song label in full-size mode, and merged with
master.  The merge is UNTESTED, because I don't have gtk 3.14 and can't see
a safe way to get it.

Moving the window: hiding window decorations definitely goes with the whole
Mini theme.  You can always move it by setting it back to full size,
dragging, and back to mini, or using whatever other move features your
window manager offers.  But if you really think it's better the other way,
I'll change it.  It doesn't directly effect me, because my window
management is weird.

Size of all text: I don't understand what you mean by this.

TingPing:
> I don't have gtk 3.14 and can't see a safe way to get it.

[JHBuild](https://wiki.gnome.org/Newcomers/BuildGnome) would be the way to do it without touching system versions.

> Moving the window: hiding window decorations definitely goes with the whole
Mini theme.  You can always move it by setting it back to full size,
dragging, and back to mini, or using whatever other move features your
window manager offers.  But if you really think it's better the other way,
I'll change it. 

I'd think using CSD would be the best option for this, keeps sane controls but can be more compact.
There has been discussion about this already but it never went anywhere.

> Size of all text: I don't understand what you mean by this.

You globally load CSS that changes every single widget in the application which changes things like the preferences dialog.

JasonLG1979:
+1 Love the idea.

But I can't get it to run though. I get this if I click the toggle.
```
    widget.show()
AttributeError: 'NoneType' object has no attribute 'show'
```
Another thought is to move it into the "Pithos" menu and give it a keyboard shortcut(if you haven't already?) The toggle just looks odd grouped with the the media control buttons. Generally in most themes that group of buttons is about the same width as the album cover if not just a little smaller. It looks weird to have be larger.

screenshot_2015-10-17_22-35-14

JasonLG1979:
Forgive the super quick mock up. It's just bits of images of other programs pasted together but if you squint a little bit and use you imagination it could work. But here's my idea of mini-mode. And really you wouldn't need to even list the station name. a dropdown arrow would be enough.
minimode

dspeyer:
Is there a line number associated with that AttributeError?

FWIW, here's a screenshot

(I am still working on this, just intermittently)

JasonLG1979:
> Is there a line number associated with that AttributeError?

The error happens while it's tryingto do this:
```python
        for widget in ui:
            if self.is_full:
                widget.show()
            else:
                widget.hide()
```


> FWIW, here's a screenshot

I get no screenshot.

JasonLG1979:
@dspeyer are you planning on moving forward with this? At this point this PR is 1 1/2 years old...

JasonLG1979:
@dspeyer I'm closing this PR.

PR #232, opened 2015-10-06
save sort preference in the stations popover dialog

kirinthos:
tracked the sort button and created a 'sort_preference' setting to track the station sort preference between successive launches.

not sure if this bothers anyone else, but I prefer to always sort A-Z


TingPing:
Need to add the default value in *PreferencesPithosDialog.py* and I'm not a fan of the name *sort_preference*. Maybe something like *sort_stations*?

kirinthos:
added the default value, and changes to sort_stations

PR #235, opened 2015-10-12
save preferences on session end (linux)

kirinthos:
Add plugin to hook into dbus for "EndSession" signal coming from org.gnome.SessionManager. Saves preferences on logout/reboot/shutdown (proper)

will have to look into other operating systems


TingPing:
I think the better solution is just saving the instant something changes. The branch [wip/gsettings](https://github.com/pithos/pithos/tree/wip/gsettings) would eventually do this for the few remaining this that save only on exit. This is just a workaround for one very specific case and doesn't solve crashes or killing the application.

PR #238, opened 2015-10-15
Get and set Explicit Content Filter State.

JasonLG1979:


TingPing:
The Pandora module is very inconsistent about it but you should probably make this into a property, at the very least combine the two setter functions.

JasonLG1979:
By inconsistent do you mean it this errors if it's called before the user is logged in? If so then yep it can't be called before the user is logged in like during the initial loading of preferences. I could throw a try in the check function so that it will silently fail and return None True or False. That way a toggle/check box in the prefs initially be set to inactive/unclick able and then if/when it returned True or False it could be set to the appropriate setting and become click able?

JasonLG1979:
*throws errors

TingPing:
> By inconsistent do you mean it this errors if it's called before the user is logged in?

No I mean't coding style in terms of using Python properties, I didn't even test this yet.

JasonLG1979:
I'll have to do some reading up on the property decorator. I pretty much know nothing about it.

PR #248, opened 2015-10-24
Create temporary local album arts files for MPRIS

### LINKED ISSUE #245: No arts in the sound menu
Hi, I am using the new 1.1.1 version on LMDE 2, MPRIS plugin is ON, so it adds the song info/controls to the system sound menu, but no album arts. To tell the truth, I do not remember how it worked with the version 0.3.18_4 on my previous system, which was LMDE 1.

dbfin:
Closes #245 

The commit consists of the following changes:

pithos.py

Note, that chronologically the code works as ENTER > 1 > (5 > 2 > 4)* > 1 > EXIT

1. Creating a temporary directory, lines 237-243, and removing it on exit, lines 1120-1125.

2. get_album_art function, lines 92-114.

    a. We accept an additional parameter tmpfile, which contains the file name and handle for the temporary file.
    b. We try to right the original image (before resizing) to the file. At first, I thought we could have done this in art_callback, but there we would have only the image 96x96, not the original 500x500.
    c. We return an additional parameter after pixbuf, which is the file name or None, depending on whether writing to the file was successful.

3. A new signal "song-art-changed", line 133, is needed to let MPRIS know when the arts for the current song was downloaded if it started playing a song before that (for example, the first song).

4. Function art_callback, lines 603-612. As was mentioned above, an additional parameter is now returned to the art_callback, so if it is not None, there is a file with the arts, and we reassign song.artFile and issue the signal, so MPRIS will know.

5. Function callback, changed lines 622-629. This function is called when the playlist is created or updated. For each new song (variable i) we first set i.artFile=None and try to create a temporary file to store the image. If the file is created successfully, as was mentioned above, we pass it to get_album_art, and if it successfully writes the image to the file, it will return the file name to art_callback, which will update song.artFile and issue the signal. Before all this happens, i.artFile should be None, as there is no file yet, should MPRIS already start playing this song.

_mpris.py

1. Line 42, connect to the new signal to know when the album arts are downloaded.

2. Line 53, use artFile instead of artUrl for the local image.

3. Lines 57-65, update the album arts link, only if it was changed for the currently playing song.

Additional comments:

- Without the new signal and signal handling, the album arts is not shown for the first song, and when one quickly switches from the 3rd song to the 5th, or from the 7th to the 9th etc. (when a new portion of the list is still being downloaded). And if we pass the name of the future file right away, it won't be updated in the sound menu even when the file is downloaded, so we need to pass None at first, and then update it using the signal.

- At first, I thought to delete the temporary image files as soon as a song finishes, but it leads to unnecessary complications. Anyway, the files are small, so even if you listen to hundred songs, the arts won't take more than several megabytes. So, we can delete them all at once in the end.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -23,6 +23,8 @@ import os
 import re
 import sys
 import time
+import shutil
+import tempfile
 import urllib.error
 import urllib.parse
 import urllib.request
@@ -87,18 +89,29 @@ class CellRendererAlbumArt(Gtk.CellRenderer):
             Gdk.cairo_set_source_pixbuf(ctx, pixbuf, x, y)
             ctx.paint()
 
-def get_album_art(url, *extra):
+def get_album_art(url, tmpfile, *extra):
     try:
         with urllib.request.urlopen(url) as f:
             image = f.read()
     except urllib.error.HTTPError:
         logging.warning('Invalid image url received')
-        return (None,) + extra
+        return (None, None,) + extra
+
+    filename = None
+    if tmpfile and tmpfile.file:
+        try:
+            tmpfile.file.write(image)
+            logging.info("Album art stored in %s"%tmpfile.name)
+            filename = tmpfile.name
+        except IOError as e:
+            logging.error("Error writing to file %s"%tmpfile.name)
+        finally:
+            tmpfile.close()
 
     with contextlib.closing(GdkPixbuf.PixbufLoader()) as loader:
         loader.set_size(ALBUM_ART_SIZE, ALBUM_ART_SIZE)
         loader.write(image)
-        return (loader.get_pixbuf(),) + extra
+        return (loader.get_pixbuf(), filename,) + extra
 
 class PlayerStatus:
   def __init__(self):
@@ -117,6 +130,7 @@ class PithosWindow(Gtk.ApplicationWindow):
     __gsignals__ = {
         "song-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "song-ended": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "song-art-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "song-rating-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "play-state-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
         "user-changed-play-state": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
@@ -219,6 +233,13 @@ class PithosWindow(Gtk.ApplicationWindow):
         aa = theme.load_icon('pithos-album-default', 128, 0)
 
         self.default_album_art = aa.scale_simple(ALBUM_ART_SIZE, ALBUM_ART_SIZE, GdkPixbuf.InterpType.BILINEAR)
+
+        try:
+            self.tempdir = tempfile.mkdtemp(prefix = 'pithos-')
+            logging.info("Created temporary directory %s"%self.tempdir)
+        except IOError as e:
+            self.tempdir = None
+            logging.error('Failed to create a temporary directory')
 
     def init_ui(self):
         GLib.set_application_name("Pithos")
@@ -580,11 +601,14 @@ class PithosWindow(Gtk.ApplicationWindow):
             return
 
         def art_callback(t):
-            pixbuf, song, index = t
+            pixbuf, filename, song, index = t
             if index<len(self.songs_model) and self.songs_model[index][0] is song: # in case the playlist has been reset
                 logging.info("Downloaded album art for %i"%song.index)
                 song.art_pixbuf = pixbuf
                 self.songs_model[index][3]=pixbuf
+                if filename:
+                    song.artFile = filename
+                    self.emit('song-art-changed', song)
                 self.update_song_row(song)
 
         def callback(l):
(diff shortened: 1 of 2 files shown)


TingPing:
So in order for this to work with the mpris extension for Gnome I have to modify it to be a `file://` url which makes sense since the metadata item is called *artUrl*. Does this change still solve the problem on whatever mpris app you are using?

dbfin:
The "file://" part itself is not that important. In fact, the patch does not even add this part, so links sent to mpris are like '/tmp/pithos-xxx/art-yyy', because I was not even sure if the mpris plugin is used on Windows, and if something like "file://c:\tmp\" would be correct on Windows.

Anyway, I believe, the main problem on my machine was that Pithos sent external "http://" links to MPRIS. In any case, what the point of asking mpris to download the file again, if you already downloaded it?

So, what the change does, it creates/removes a temporary directory, where all album arts files are temporarily stored. The files stored there are original (not resized), 500x500. This improves the quality of the picture shown in the sound menu. If a song starts playing when the album arts is not available yet, we send a signal to inform MPRIS once it is downloaded.

TingPing:
> In fact, the patch does not even add this part

I know.. that is why I said you need to add it.

> The "file://" part itself is not that important. 

Except the specification says it requires it and the implementation I use requires it.

dbfin:
Yes, sure. Regarding the Windows, I believe the mpris plugin is irrelevant for Windows, so I will add file:// inside the plugin rather than Pithos itself. This way both will have links according to specifications.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   From: TingPingSent: Saturday, October 24, 2015 2:27 PMTo: pithos/pithosReply To: pithos/pithosCc: VadimSubject: Re: [pithos] Create temporary local album arts files for MPRIS (#248)
In fact, the patch does not even add this part


I know.. that is why I said you need to add it.


The "file://" part itself is not that important. 


Except the specification says it requires it and the implementation I use requires it.

—Reply to this email directly or view it on GitHub.

dbfin:
And sorry, I just misunderstood what you were saying.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    From: TingPingSent: Saturday, October 24, 2015 2:27 PMTo: pithos/pithosReply To: pithos/pithosCc: VadimSubject: Re: [pithos] Create temporary local album arts files for MPRIS (#248)
In fact, the patch does not even add this part


I know.. that is why I said you need to add it.


The "file://" part itself is not that important. 


Except the specification says it requires it and the implementation I use requires it.

—Reply to this email directly or view it on GitHub.

TingPing:
@dbfin Site note, please don't reply to github via email, it just breaks formatting.

dbfin:
I'm here (on the website). The formatting looks OK to me, so, just out of curiosity, what do you mean? But, anyway, I will try to respond on the website, if it is more convenient.

TingPing:
> The formatting looks OK to me, so, just out of curiosity, what do you mean? 

Because I have been editing every message :P

dbfin:
Got it, thanks! I, actually, thought GitHub removes all irrelevant parts such as the original message quoted below my reply. Was this the problem?

TingPing:
Yes the quote and everything was being included. I generally also suggest avoiding email because users can modify their comment and emails won't see those changes, I happen to do this very often. Also if you ever use formatting such as your original comment using numbered lists you can see a preview of it rendered on the webui

dbfin:
Makes sense. Thanks for the advise. I believe it is GitHub's fault that they offer the convenience of replying by email, but post everything including the original conversation at the end of the reply. Is there some place to report the issue?

dbfin:
I submitted the change. It may seem a bit more complicated than just adding "file://" in front of the artFile, but this way we protect it from having double protocol scheme specifications such as "file://file://" or "file://http://", should something change in the future.

Besides, on Linux the path to the file sent to the plugin takes form "/tmp/..." so we should add "file://" to it, while on Windows (even though I am not sure it is used on Windows at all, but in any case) it would be something like "c:\tmp\...", so, according to [specifications](http://www.ietf.org/rfc/rfc1738.txt), see 3.10, we should add "file:///" to indicate the empty host. Even on Linux, something may change in the future, and, for example, paths without the leading slash might be sent to the plugin. So, this seems to be a more robust approach to handle local file names.

TingPing:
We don't need any of that complication.. just prepend file:// in `get_album_art()`... and I don't know why you are bringing up Python 2 which is completely unsupported.

dbfin:
Regarding Python 2.x, I remember compiling Pithos using it, but if it is not supported anymore, this makes the life easier.

Regarding adding 'file://', I would certainly keep it within the mpris plugin, as it is mpris that requires this part, which is otherwise unnecessary. Moreover, I am not even sure that regular OS operations would work with this 'file://' part. So, it makes sense to add this part for mpris only, to avoid bugs in the main part.

So, the questions is whether to keep the following:

1. Checking whether there is already a protocol in the file name.

2. Checking if the file name starts with the slash. Can this plugin be used on a system that does not use leading slashes (such as Windows, or any other). I am not sure what names are used on OSX, I would guess same as on Linux, but not sure.

dbfin:
@TingPing I have already removed Python 2.x support. What about the points 1 and 2 above. It is literally 4 additional lines of code, but makes it foolproof against many possible scenarios and future bugs.

TingPing:
> Regarding adding 'file://', I would certainly keep it within the mpris plugin, as it is mpris that requires this part, which is otherwise unnecessary. Moreover, I am not even sure that regular OS operations would work with this 'file://' part. So, it makes sense to add this part for mpris only, to avoid bugs in the main part.

Who cares what regular OS operations work with, just document that the album art property is a URI and will always be a URI. This entire PR is specific to the mpris plugin, moving like 10% of the logic out of the core doesn't make anything cleaner than just setting a proper URI from the start.

> Checking if the file name starts with the slash. Can this plugin be used on a system that does not use leading slashes (such as Windows, or any other). I am not sure what names are used on OSX, I would guess same as on Linux, but not sure.

The python API you use for a tempdir should always return an absolute path. You keep talking about other platforms for some reason, file:// is a valid URI everywhere...

dbfin:
OK, I see where the misunderstanding comes from. 'file:///' is a valid URI everywhere, so '/tmp/' should become 'file:///tmp/', so we add 'file://', and 'tmp/' or 'c:/tmp/' should become 'file:///tmp/' or 'file:///c:/tmp/, so we add 'file:///' instead. This is according to the specification I linked to. This is why my question.

TingPing:
If you want to always be technically correct:
```python
urllib.parse.urljoin('file://', urllib.parse.quote(filename))
```

dbfin:
Perfect, I will look into this, and, will move the 'file://' part to pithos.py, as you suggested. Anyway, it is a very minor part of the patch. Thanks.

dbfin:
OK, I reverted the last changes, and did according to your comments. Now it is indeed much easier.

PR #255, opened 2015-12-11
Better logging.

JasonLG1979:
Doesn't change console logging behavior. 
Creates a ~/.pithos folder and logs at a WARNING level to 'event.log' by default. 
added new command line args -f and -e.
-f sets the logging level for the log file to INFO and -e sets the logging level for the log file to DEBUG.
Should make debugging and trouble shooting user issues easier.

### DIFF
--- a/pithos/application.py
+++ b/pithos/application.py
@@ -13,7 +13,7 @@
 #You should have received a copy of the GNU General Public License along
 #with this program.  If not, see <http://www.gnu.org/licenses/>.
 ### END LICENSE
-
+import os
 import sys
 import signal
 import logging
@@ -40,6 +40,10 @@ class PithosApplication(Gtk.Application):
                              _('Show debug messages'), None)
         self.add_main_option('test', ord('t'), GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
                              _('Use a mock service instead of connecting to the real Pandora server'), None)
+        self.add_main_option('verbose_file', ord('f'), GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
+                             _('log info messages to file'), None)
+        self.add_main_option('debug_file', ord('e'), GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
+                             _('log debug messages to file'), None)
 
     def do_startup(self):
         Gtk.Application.do_startup(self)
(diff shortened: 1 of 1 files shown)


TingPing:
Maybe this is valuable on Windows (which we still don't support), but uh, what is wrong with `pithos -v 2> event.log`?

JasonLG1979:
Because this outputs to a log file at a WARNING level all the time. Users don't generally run Pithos from the terminal so this way when something goes wrong they/we would have a record of it.

JasonLG1979:
Like I said it would help with trouble shooting user issues. cat .pithos/event.log is a lot easier then running Pithos over and over again from the terminal to find sporadic/random errors. How many times have you tried to trouble shoot user issues here on github and wished you could get better info from them?

TingPing:
Personally I'd say it has been pretty easy to tell users "rerun with `pithos -v` and pastebin the output.".

JasonLG1979:
> Personally I'd say it has been pretty easy to tell users "rerun with pithos -v and pastebin the output.".

Only helps after the fact. And if the error doesn't happen again you have no idea what happened the 1st time. Like I said this outputs at a WARNING level all the time. Errors and such get logged to file no matter if you run Pithos from the terminal or not.

JasonLG1979:
If you'd like to continue to trouble shoot "This happened, I have no idea why and I have no idea how to get you more info" issues. Then I guess that's your call,lol!!!

TingPing:
Heh, I come from HexChat which is 60k lines of C with absolutely no logging of any kind, debugging Pithos is already a dream compared to this :P

JasonLG1979:
> Heh, I come from HexChat which is 60k lines of C with absolutely no logging of any kind, debugging Pithos is already a dream compared to this :P

For you maybe but not the average user.

JasonLG1979:
I guess I don't understand the resistance. You're acting like I'm from Mars because I'd like to log errors to a log file,lol!!! It's a pretty common practice...

TingPing:
Well I just like things that are added to have very clear and obvious 'value' to me. Features added just because don't lead to a clean codebase. I know that this is a very minor patch and I am not necessarily that against it I just don't think it is useful.

The one and only real problem I can think of it atm is that it will just continually grow the file forever since it doesn't seem to limit the size or even rotate it.

JasonLG1979:
> The only real problem I can think of it atm is that it will just continually grow the file forever since it doesn't seem to limit the size or even rotate it. Oh and it puts it in a terrible location.

It should log very little to nothing by default. And I don't know the proper location. It's user specific so somewhere in their home dir?

JasonLG1979:
Best I can come up with to limit log file size is something like this:
```python
    if os.path.exists(log_file):
        line_offset = []
        offset = 0
        number_of_lines = 0
        with open(log_file, 'rb+') as log:
            for line in log:
                #get the number of lines and each line's offset
                line_offset.append(offset)
                offset += len(line)
                number_of_lines += 1
            if number_of_lines > 250: #250 lines is plenty enough for debugging problems
                delete_before_line = number_of_lines - 250 #the line we want to seek to
                log.seek(line_offset[delete_before_line]) #seek to the line offset of the delete_before_line 
                keeper_lines = log.read() #the last 250 lines
                log.seek(0) #rewind to start of file
                log.write(keeper_lines) #write back the last 250 lines 
                log.truncate()
            log.close()
```
This would keep the last 250 log messages. 250 lines would be a pretty long log but the file size as far as space would be pretty tiny.

Issue #268, opened 2016-02-20
Proper(ish) MPRIS implementation


JasonLG1979:
It's known that MPRIS in Pithos is broken/not implemented properly. It's also apparent from skimming the MPRIS spec that it really wasn't designed for a "radio" app like Pithos. It's geared more towards your traditional media players in a lot of respects.(an example is that a media player is expected to have a "stop" function/button, which Pithos lacks.)

Aside from fixing what we do have(which is definitely needed) MPRIS does offer a few more neat features we could implement like track time/progress indication, and playlist support.

I opened this issue in the hopes of sparking discussion and to track progress on the way to a properly working MPRIS implementation in Pithos. And also as a way to wrangle MPRIS bugs.

TingPing:
I don't think what we have is *that* broken but yes it could use improvements.

> we could implement like track time/progress indication

:+1: 

> and playlist support.

Playlists don't map perfectly to stations, but it is an interesting idea to play with.

JasonLG1979:
> I don't think what we have is that broken but yes it could use improvements.
It doesn't "appear" broken but it is. It would be nice if it worked properly.
 
> Playlists don't map perfectly to stations, but it is an interesting idea to play with.

That's where the "ish" part comes in. My thought is for stations to appear as empty playlists(if that's possible?) When you click on the station the playlist would populate as you'd expect.

TingPing:
> My thought is for stations to appear as empty playlists(if that's possible?) When you click on the station the playlist would populate as you'd expect.

That is what I was thinking.

> It doesn't "appear" broken but it is. It would be nice if it worked properly.

Other than not showing the length.. honestly it works perfectly for me.

JasonLG1979:
> Other than not showing the length.. honestly it works perfectly for me.

It works for the most part but it's flaky at best. On every widget(or what ever you call them) I tested on, GNOME, Unity, and Cinnamon something always breaks after a while. On KDE 5 + it doesn't work at all.

JasonLG1979:
So maybe we could set a roadmap for MPRIS? My thoughts are...

1. Clean up what we have and make it as close to spec as makes sense.
2. Add track time/progress indication.
3. Play with the whole playlist idea.

I'll have a week or so off because I'm transitioning to a new job, so I'll read up on the spec and look at a few examples from other apps to see what I come up with. After that the track time/progress indication stuff is pretty easy I played with that and got it working a couple months ago it's just a matter of remembering what I did,lol!!!

[...]

Issue #261, opened 2015-12-21
MPRIS metadata is missing the mpris:length field


maddin77:
Hi,

it seems like the `mpris:length` metadata field is missing, which means i'm not able to find out the total duration of the current song.

Is there another way to get the duration? Or if not, is that something that could be added in the future?

TingPing:
I think @JasonLG1979 brought this up in a previous issue. Would be a good addition.

JasonLG1979:
The problem is that MPRIS expects the track length when the rest of the meta data is loaded but we don't know it yet at that point. A work around would be to use the duration provided by Pandora but it's not very accurate.

JasonLG1979:
After testing it on various applets, even if I get the length from Pandora so it is known at song change just like the rest of the metadata, It's pretty dang flaky.
 
Ubuntu's sound menu of course doesn't support it, so that's that.

 The Media player indicator GNOME extension will show it if you toggle the setting on then off again.

The default Cinnamon sound applet doesn't show it and has no such toggle.

The only applet I can get to work correctly is:

https://github.com/Koutch/Cinnamon-sound-with-apps-volume

I started an issue with Media player indicator to try to find out if it's a bug on our side or theirs.

https://github.com/eonpatapon/gnome-shell-extensions-mediaplayer/issues/246

JasonLG1979:
@maddin77 give this branch a try. IMHO it vastly improves MPRIS support Pithos, including providing mpris:length.

maddin77:
Yes, that seems to work as expected. Thanks.

I also tried the GNOME mediaplayer extension ([r545-1 from the AUR](https://aur.archlinux.org/packages/gnome-shell-extension-mediaplayer-git/)), and i can't seem to reproduce the issue described in eonpatapon/gnome-shell-extensions-mediaplayer#246. The position slider shows up and seems to display the correct value for me.

JasonLG1979:
> I also tried the GNOME mediaplayer extension (r545-1 from the AUR), and i can't seem to reproduce the issue described in eonpatapon/gnome-shell-extensions-mediaplayer#246. The position slider shows up and seems to display the correct value for me.

With which branch?

JasonLG1979:
@TingPing you could maybe merge this issue with #268 maybe?

maddin77:
> With which branch?

Pithos: https://github.com/JasonLG1979/pithos/tree/MPRIS_fixes
[...]

PR #269, opened 2016-02-26
Various MPRIS fixes.

JasonLG1979:
General clean up.
New signals specifically for MPRIS.
Fixed volume slider scaling in sound applets.
Fixed volume sliders not staying in sync.
Mapped Pithos ratings to something MPRIS understands.
Added Position support.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -127,10 +127,12 @@ class PithosWindow(Gtk.ApplicationWindow):
     __gsignals__ = {
         "song-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "song-ended": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
-        "song-art-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
-        "song-rating-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "play-state-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
         "user-changed-play-state": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
+        "metadata-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "volume-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "playlist-cleared": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
+        "sync-position": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
     }
 
     volume = GtkTemplate.Child()
@@ -400,6 +402,7 @@ class PithosWindow(Gtk.ApplicationWindow):
                     self.current_song_index = None
                     self.songs_model.clear()
                     self.get_playlist(start = True)
+                    self.emit('playlist-cleared', True)
 
             if self.filter_state is not None and self.filter_state != current_checkbox_state:
                 self.worker_run(set_content_filter, (current_checkbox_state, ), get_new_playlist)
@@ -570,6 +573,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.set_title("Pithos - %s by %s" % (self.current_song.title, self.current_song.artist))
 
         self.emit('song-changed', self.current_song)
+        self.emit('metadata-changed', self.current_song)
 
     @GtkTemplate.Callback
     def next_song(self, *ignore):
@@ -648,7 +652,7 @@ class PithosWindow(Gtk.ApplicationWindow):
                 self.songs_model[index][3]=pixbuf
                 if file_url:
                     song.artUrl = file_url
-                    self.emit('song-art-changed', song)
+                    self.emit('metadata-changed', song)
                 self.update_song_row(song)
 
         def callback(l):
@@ -722,6 +726,7 @@ class PithosWindow(Gtk.ApplicationWindow):
             self.stop()
             self.current_song_index = None
             self.songs_model.clear()
+            self.emit('playlist-cleared', True)
         logging.info("Selecting station %s; total = %i" % (station.id, len(self.stations_model)))
         self.current_station_id = station.id
         self.current_station = station
@@ -750,6 +755,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.current_song.duration = self.query_duration()
         self.current_song.duration_message = self.format_time(self.current_song.duration)
         self.check_if_song_is_ad()
+        self.emit('metadata-changed', self.current_song)
         self.player_status.pending_duration_query = False
 
     def on_gst_duration_changed(self, bus, message):
@@ -757,6 +763,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.current_song.duration = self.query_duration()
         self.current_song.duration_message = self.format_time(self.current_song.duration)
         self.check_if_song_is_ad()
+        self.emit('metadata-changed', self.current_song)
       else:
         self.player_status.pending_duration_query = True
 
@@ -862,6 +869,9 @@ class PithosWindow(Gtk.ApplicationWindow):
             else:
                 logging.debug("Buffer recovery. User paused")
             self.player_status.began_buffering = None
+            #Tell everyone to update their clocks after we're done buffering or 
+            #in case it takes a while after the song started signal for actual playback to begin.
+            self.emit('sync-position', self.query_position() or 0)
         self.player_status.buffer_percent = percent
         self.update_song_row()
 
@@ -870,6 +880,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         scaled_volume = math.pow(volume, 1.0/3.0)
         self.volume.handler_block_by_func(self.on_volume_change_event)
         self.volume.set_property("value", scaled_volume)
+        self.emit('volume-changed', scaled_volume)
         self.volume.handler_unblock_by_func(self.on_volume_change_event)
 
     def on_gst_volume(self, player, volumespec):
@@ -977,7 +988,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         song = song or self.current_song
         def callback(l):
             self.update_song_row(song)
-            self.emit('song-rating-changed', song)
+            self.emit('metadata-changed', song)
         self.worker_run(song.rate, (RATE_LOVE,), callback, "Loving song...")
 
 
@@ -985,7 +996,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         song = song or self.current_song
         def callback(l):
             self.update_song_row(song)
-            self.emit('song-rating-changed', song)
+            self.emit('metadata-changed', song)
         self.worker_run(song.rate, (RATE_BAN,), callback, "Banning song...")
         if song is self.current_song:
             self.next_song()
@@ -994,14 +1005,14 @@ class PithosWindow(Gtk.ApplicationWindow):
         song = song or self.current_song
         def callback(l):
             self.update_song_row(song)
-            self.emit('song-rating-changed', song)
+            self.emit('metadata-changed', song)
         self.worker_run(song.rate, (RATE_NONE,), callback, "Removing song rating...")
 
     def tired_song(self, *ignore, song=None):
         song = song or self.current_song
         def callback(l):
             self.update_song_row(song)
-            self.emit('song-rating-changed', song)
+            self.emit('metadata-changed', song)
         self.worker_run(song.set_tired, (), callback, "Putting song on shelf...")
         if song is self.current_song:
             self.next_song()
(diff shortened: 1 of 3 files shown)


pithos/plugins/_mpris.py
```diff
@@ -158,20 +189,26 @@ def GetAll(self, interface_name):
...
+                'CanGoNext': True,
                 'CanGoPrevious': False,
-                'CanPlay': self.window.current_song is not None,
-                'CanPause': self.window.current_song is not None,
-                'CanSeek': False,
+                'CanPlay': True,
+                'CanPause': True,
+                #CanSeek has to be True for some sound applets
+                #to show the song position/duration info
+                'CanSeek': True,
```
TingPing: I get why you do it, but as I've mentioned before I don't want our client to lie to fix other projects. Thankfully everything is open an can be fixed.

JasonLG1979: I agree. I'll see if I can get some traction on that issue. Until it's fixed though?

pithos/pithos.py
```diff
@@ -127,10 +127,11 @@ class PithosWindow(Gtk.ApplicationWindow):
     __gsignals__ = {
         "song-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "song-ended": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
-        "song-art-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
-        "song-rating-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "play-state-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
         "user-changed-play-state": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
+        "mpris-metadata-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "mpris-play-state-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "mpris-volume-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
```
TingPing: I don't like these `mpris` prefixes, signals shouldn't be plugin specific. I also don't like the duplicated play-state signals.

pithos/plugins/_mpris.py
TingPing: Well I am not happy about it.. but I suppose we can. You will need to actually implement `Seek()` and `SetPosition()` though (it would do nothing, but we shouldn't error when it is called...)

pithos/pithos.py
JasonLG1979: the mpris prefix is so you know what it's for otherwise someone unfamiliar with the code would not know what the signal is for.

The duplicate signal is because the main play-state-changed signal is a bool mpris needs either 'Playing', 'Paused' or 'Stopped'.

pithos/plugins/_mpris.py
JasonLG1979: I can do that.

JasonLG1979: Done.

JasonLG1979:
Amended with references to known bugs in clients mentioned in #268.

JasonLG1979:
@TingPing I'm done with this(as far as I can tell?) can you take a look at it. As soon as this is done I'd like to start on adding PlayList(stations) support. Thank :)

JasonLG1979:
Amended to fix the track_id problem mentioned in #261 and to add a custom 'RateSong' Method.

pithos/pithos.py
JasonLG1979: Removed the mpris prefixes, and removed the duplicated play-state, as it's not rally needed now.
https://github.com/johnhoran/Laine/issues/27

TingPing:
@JasonLG1979 Just an update on why I am slow to merge this; Lately my schedule has been a bit messed up and due to RSI pains in my wrist I can barely type and on top of that you seem to still be tweaking things and working with upstream so I saw no reason to rush.

JasonLG1979:
No problem. I also would like for some of the upstream stuff to be cleared out. Although I'm sure how far we're going to get with the (IMHO) bugs in GNOME media extension.

JasonLG1979:
Like I said in that Laine issue. It's a pain in the butt when you have a somewhat ambiguous spec and no real reference implementation. You end up with everyone interpreting the spec a little differently, which kinda defeats the purpose of a spec in the 1st place, lol!!!

TingPing:
@JasonLG1979 BTW Gnome 3.20 released with built in MPRIS support: https://help.gnome.org/misc/release-notes/3.20/

JasonLG1979:
> BTW Gnome 3.20 released with built in MPRIS support:

Cool. Very limited support(pretty much the equivalent of the Ubuntu sound applet.) It should work fine with it.

JasonLG1979:
@TingPing as soon as/if you merge this and my MPRIS pull request I've got another ready that makes Pithos use symbolic icons all around.(MPRIS, notifications, and ratings.)

JasonLG1979:
screenshot from 2016-04-01 14-56-02

pithos/pithos.py
```diff
@@ -127,10 +127,11 @@ class PithosWindow(Gtk.ApplicationWindow):
     __gsignals__ = {
         "song-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "song-ended": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
-        "song-art-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
-        "song-rating-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "play-state-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
         "user-changed-play-state": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
+        "metadata-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "volume-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "songs_model_cleared": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
```
TingPing: Maybe call this `playlist-cleared`?

JasonLG1979: That makes sense. Will do.

JasonLG1979: Done.

pithos/plugins/_mpris.py
```diff
@@ -31,55 +35,86 @@ def __init__(self, window):
...
+        random_string = ''.join(random.SystemRandom().choice(string.ascii_letters + string.digits) for x in range(56))
+        self._unique_trackid = ('/org/mpris/MediaPlayer2/TrackList/%s%s' %(song.index, random_string))
+        self._valid_trackids.append(self._unique_trackid)
+        logging.info("MPRIS trackid for %s by %s: %s"%(song.title, song.artist, self._unique_trackid))
+
+    def _song_index_from_trackid(self, trackid):
+        """Extracts the song.index from a trackid by slicing the object path
+        from the begining and the random numbers from the end.
+        """
+        return int(trackid[34:-56])
```
TingPing: I'm not sure I like hardcoding these indexes, what is the string you are expecting here?

JasonLG1979: The string is the MPRIS trackId. It contains the required '/org/mpris/MediaPlayer2/TrackList/' + song_index + a random string of letters and numbers. The trackid must be unique as per the MPRIS docs. It's how MPRIS clients refer to the song. For example ' /org/mpris/MediaPlayer2/TrackList/1D5VDR0XZ6PDlfoD4QppsZesf6AXOjK0hCLzVDPIgHqARoAXcPn3Ekw49'. self._song_index_from_trackid takes the trackid and turns it back to a song index for the purpose of rating the song.

JasonLG1979: The trackId is also used through out MPRIS for it's playlist and tracklist interfaces and to tell if the song has changed.

pithos/plugins/_mpris.py
```diff
@@ -110,19 +175,40 @@ def _get_playback_status(self):
...
+        if self._metadata:
+            return self._metadata
+        else:
+            return {"mpris:trackid": '/org/mpris/MediaPlayer2/TrackList/NoTrack',
+                    #workaround for https://github.com/eonpatapon/gnome-shell-extensions-mediaplayer/issues/247
+                    "xesam:url": "",}
+    @property
+    def _current_volume(self):
+        volume = self.window.player.get_property("volume")
+        scaled_volume = math.pow(volume, 1.0/3.0)
```
TingPing: This doesn't scale the same as normal volume.

pithos/plugins/_mpris.py
JasonLG1979: self._valid_trackids is a list of valid trackid's. They are checked against in the rating method and cleared on the 'playlist-cleared' signal.

TingPing: Maybe I am blind, but where is RateSong in the spec?

pithos/plugins/_mpris.py
JasonLG1979: Just piping the volume through to MPRIS results in incorrect volume scaling. I'm not sure why but this fixes that.

JasonLG1979: If I had to guess it's because we do the opposite in the main ui. (math.pow(volume, 3.0/1.0)

JasonLG1979: The end result is that the sliders in the main UI and MPRIS are in sync with this.

TingPing:
Ok so I am going to remove RateSong() for this first revision. I also am using `song.trackToken` for the trackid.

JasonLG1979:
1 last little tweak. Added a sync-position signal. MPRIS uses it's own clock and doesn't know when/if Pithos is buffering. As a result it's clock will keep running when Pithos is buffering and go out of sync this will fix that.

JasonLG1979:
> Ok so I am going to remove RateSong() for this first revision. I also am using song.trackToken for the trackid.

You can't song.trackToken contains invaild chars sometimes and will throw exceptions.

TingPing:
> You can't song.trackToken contains invaild chars sometimes and will throw exceptions.

We can throw it through base64 if we must.

JasonLG1979:
> We can throw it through base64 if we must.

I don't know what that will do but as long as this is achieved it'll do.
https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-marshaling-object-path

TingPing:
Ok, well we can just leave that out this revision since we don't use it yet.

JasonLG1979:
Songs HAVE to have a trackId...

JasonLG1979:
https://specifications.freedesktop.org/mpris-spec/latest/Track_List_Interface.html#Mapping:Metadata_Map

"The mpris:trackid attribute must always be present, and must be of D-Bus type "o". This contains a D-Bus path that uniquely identifies the track within the scope of the playlist. There may or may not be an actual D-Bus object at that path; this specification says nothing about what interfaces such an object may implement."

TingPing:
Oh we just never did before :P

JasonLG1979:
I know. I really wish there was a reference implementation. I tried to follow the spec as best I could. But since there is no reference implementation everyone seems to have their own little twist on it(hence all the workarounds)

TingPing:
Ok so I'm just going to use `song.trackToken` sha1 hashed, should be fine, only ascii. Hurry up and push whatever else you wanted to finish (as a new commit), so we can finally merge this :P

JasonLG1979:
But since we clear the valid_trackids list when the playlist clears we might be able to get away with just using '/org/mpris/MediaPlayer2/TrackList/' + song_index

JasonLG1979:
Whatever we use it has to start with '/org/mpris/MediaPlayer2/TrackList/' otherwise it's not a valid path.

TingPing:
> Whatever we use it has to start with '/org/mpris/MediaPlayer2/TrackList/' otherwise it's not a valid path.

This is false, re-read the spec:

> Media players may not use any paths starting with /org/mpris unless explicitly allowed by this specification. Such paths are intended to have special meaning, such as /org/mpris/MediaPlayer2/TrackList/NoTrack to indicate "no track".

JasonLG1979:
Aside from what you want to do about the trackid I'm done do what you will,lol!!!

JasonLG1979:
> This is false, re-read the spec:

Then what? '/org/pithos/TrackList/'

TingPing:
`/io/github/Pithos/TrackId/bc491a2d70035349e75bb719070a2823a900c788` is what I used for now.

JasonLG1979:
>/io/github/Pithos/TrackId/bc491a2d70035349e75bb719070a2823a900c788 is what I used for now.
They have to be unique so /io/github/Pithos/TrackId/ + whatever unique stuff for every song

TingPing:
Ofc was just an example of a song.

JasonLG1979:
As long as it works I'm not married to any naming scheme.

JasonLG1979:
Merge what you want of it and I'll test it out.

JasonLG1979:
My plan is this then the new default cover pr, then the additional audio urls then the symbolic icons thing then fix the settings not actually applying when you click apply in the prefs. It would be nice if we could do a release for the upcoming Ubuntu and Fedora releases.

PR #276, opened 2016-03-06
New default background icon

JasonLG1979:
Cleaner look.
Dynamically created default album cover.
Dynamically change default album cover in main ui and MPRIS with theme change.
Changed the rate background to svg.
Fixed icon centering on rate background.
Removed deprecated and unnecessary things in CellRenderers.

### DIFF
--- a/pithos/pandora/fake.py
+++ b/pithos/pandora/fake.py
@@ -113,15 +113,22 @@ class FakePandora(Pandora):
             'artistName':"ArtistName",
             'audioUrlMap': {
                 'highQuality': {
+                    'encoding': 'aac',
+                    'bitrate': '32',
                     'audioUrl': audio_url
                 },
                 'mediumQuality': {
+                    'encoding': 'aac',
+                    'bitrate': '32',
                     'audioUrl': audio_url
                 },
                 'lowQuality': {
+                    'encoding': 'aac',
+                    'bitrate': '32',
                     'audioUrl': audio_url
                 },
             },
+            'trackLength':121,
             'trackGain':0,
             'trackToken':'5908540384',
             'songRating': 1 if c%3 == 0 else 0,


JasonLG1979:
screenshot from 2016-03-06 09-55-09

pithos/pithos.py
```diff
@@ -228,7 +228,7 @@ def init_core(self):
         self.art_worker = GObjectWorker()
 
         theme = Gtk.IconTheme.get_default()
-        aa = theme.load_icon('pithos-album-default', 128, 0)
+        aa = theme.load_icon('pithos-album-default', ALBUM_ART_SIZE, 0)
 
         self.default_album_art = aa.scale_simple(ALBUM_ART_SIZE, ALBUM_ART_SIZE, GdkPixbuf.InterpType.BILINEAR)
```
TingPing: If it is the correct size at load this scaler can be removed?

JasonLG1979: Yes after playing with it the scaler can be removed. I've also been playing with a little padding and a subtle shadow.

What do you think of this? If you like it I can amend this commit with the changes.
screenshot from 2016-03-06 14-58-54

TingPing:
Too much padding for my tastes.

JasonLG1979:
That can be adjusted. What do you think of the idea of it though?

JasonLG1979:
Better?
screenshot from 2016-03-06 16-38-28

TingPing:
Certainly better but I'm not sure I prefer it overall, keep it to another PR at least.

JasonLG1979:
OK

JasonLG1979:
>If it is the correct size at load this scaler can be removed?

Fixed.

JasonLG1979:
I also "pixel-fit" the default cover it now looks like this. Very minor change but it looks a tiny bit cleaner.
pithos-album-default

http://dcurt.is/pixel-fitting

JasonLG1979:
Amended to fix icons not centering properly when Gtk.IconSize.MENU doesn't return a 16x16px icon. And it now uses a svg for the rate background.

TingPing:
Do you think making a symbolic variant would be a good idea? Would fit in better when the dark theme is used for example.

JasonLG1979:
I think making our own symbolic icons would not be a bad idea. We could get rid of the white gradient that way. It looks tacky IMHO,lol!!!

JasonLG1979:
I'll do some mockups based on the default high contrast theme and see what I can come up with.

JasonLG1979:
Off the top of my head...

icon mockups

JasonLG1979:
To me the shadow looks tacky but it is necessary to make sure the icon stands out against whatever colors are on the album cover. The problem gets worse with symbolic icons as they are monochrome and often black or white. You would still need the background but you'd have to make it symbolic to and inverted to the icon. (white icon + white bg = white smudge, black icon + black bg = black smudge) The best solution is to get the icon off the cover, some where we don't have to worry about what color the background is so we don't have to shadow it.

JasonLG1979:
Another option is to scrap the icons all together and just use text. Something like this:
screenshot from 2016-03-25 16-17-32

TingPing:
I prefer the icons to text. I was mostly talking about the album art having a symbolic version though.

JasonLG1979:
The problem with that is that the default cover is not rendered dynamically. It's assigned(?) once statically in self.default_album_art and rendered as an image.

JasonLG1979:
There's also the fact that symbolic icons aren't meant to be large icons. They're meant to be used as emblems, buttons, menu items and such. Really if we want to use a symbolic icon for the default icon we should just put a 1 px frame around the cellrender and use audio-x-generic-symbolic.

JasonLG1979:
Another option is to not even have a default album cover. Just render a border and a audio-x-generic-symbolic icon.
screenshot from 2016-03-26 11-55-49
screenshot from 2016-03-26 11-57-09
screenshot from 2016-03-26 11-57-37
screenshot from 2016-03-26 12-05-32

JasonLG1979:
@TingPing, what do you think of this. I got rid of the default album cover. Now a svg is defined at render time that chages color based on the background and foreground of the treeview, then if there is no cover a audio-x-generic-symbolic is overlayed on the dynamically created background. Basically the "default album cover" changes to match the window and icon theme.

TingPing:
> Dynamically created default album cover

Gtk already supports symbolic icon coloring so we shouldn't need to do extra work, probably see how other symbolic icons are already used.

JasonLG1979:
> I guess you are using Gtk for coloring the main icon, so perhaps that is all you can do.

I tried to use https://people.gnome.org/~shaunm/girdoc/C/Gtk.IconInfo.load_symbolic.html to make sure the actual symbolic icon stood out on the background but I couldn't get it to work. I'm sure it was something I was doing wrong...

TingPing:
I have no clue where you found those docs, but use for future reference: https://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/IconInfo.html#Gtk.IconInfo.load_symbolic

JasonLG1979:
I figured it out. Without that function the there's a chance that the symbolic icon doesn't stand out against the background.... Checkout the latest amend. Did a little bit of coloring and changing up on the ratings icons.

TingPing:
- Really like using *audio-x-generic-symbolic*
- Not a fan of the coloring the rating icons
- On the fence about the border

JasonLG1979:
The border is no problem to change although I'd still like a border if there is no cover.

The icons are no big deal I can change them back to the way they were.

TingPing:
Also the icons should be cached since they rarely change (and just cleared on theme changes).

JasonLG1979:
> Also the icons should be cached since they rarely change (and just cleared on theme changes)

You can't cache the icons unless you connect with some kind of signal for a theme change. That would break the whole effect. That fact that they are not cached is what makes the whole thing change to fit the theme.

JasonLG1979:
There. Changed the rate icons back and got rid of the border/frame. Better?
screenshot from 2016-03-27 16-28-33

JasonLG1979:
Turn out SytleContext does have a changed signal I'll see about using that to update the icons.

JasonLG1979:
Yep, I've managed to wire it to a handler and it fires on window and icon theme changes. It's just a matter of wiring it to a method that updates the icon cache for the cellrender and we're set.

JasonLG1979:
@TingPing - There, now the icons are cached and only updated in the event of a window/icon theme change.

What do you think about making the rating icons symbolic? If I can figure out how to make the rate_bg symbolic I can set it as the bg color and the rating icon as the fg color?

TingPing:
> What do you think about making the rating icons symbolic? If I can figure out how to make the rate_bg symbolic I can set it as the bg color and the rating icon as the fg color?

I'm not against it but I didn't love what you mocked up, lets keep that for another issue and get this merged first =)

JasonLG1979:
Right on. The only real part I don't like is:
```
        background = (('<svg width="96" height="96"><rect width="96" height="96" ') +
                      ('style="fill:%s;stroke-width:2;stroke:%s" /></svg>'
                       %(self.get_hex(bg_color), self.get_hex(fg_color))))
```
It's pretty ugly. Some pointers to make it look nicer would be great, thanks.

JasonLG1979:
> I didn't love what you mocked up

You mean the headerbar mockup?

JasonLG1979:
This looks nicer and works. What do you think?
```
        background = '''
        <svg width="96" height="96">
        <rect width="96" height="96"style="fill:{bg};stroke-width:2;stroke:{fg}" /></svg>
        '''.format(bg=self.get_hex(bg_color), fg=self.get_hex(fg_color))
```

JasonLG1979:
I'll save it for another PR but I wanted to get your thoughts on this. Symbolic icons for the ratings. It would adapt to the themes just like the default background/cover.
screenshot from 2016-03-29 14-34-03
screenshot from 2016-03-29 14-36-35
screenshot from 2016-03-29 14-37-50
screenshot from 2016-03-29 14-39-24

JasonLG1979:
@TingPing This now?

JasonLG1979:
This branch now has a lot of conflicts. It'll be easier for me to roll it back and start over from the current state of Pithos Master than to fix all of them. As I've said before my long game is to have Pithos use symbolic icons for ratings, and notifications and MPRIS(if there is no cover from Pandora). Something like this:

stuff

PR #283, opened 2016-04-03
Use additionalAudioUrl

JasonLG1979:
New audio quality map:
Free users:
high = 128 mp3
Pandora One users:
high = 192 mp3
Everyone:
med = 64 aacplus
low = 32 aacplus

Advantages:
Free users get a 128 mp3 stream for high.
Pandora One users get a low quality.

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -324,7 +324,11 @@ class Station:
 
     def get_playlist(self):
         logging.info("pandora: Get Playlist")
-        playlist = self.pandora.json_call('station.getPlaylist', {'stationToken': self.idToken, 'includeTrackLength': True}, https=True)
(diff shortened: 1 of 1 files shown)


TingPing:
By the way the only reason I didn't merge this is because I didn't like the hacky ad checking work around for when adts is used. I don't know of an alternative I just don't want to add a feature to only have to workaround more. That said it still might be worth merging just putting it out there.

JasonLG1979:
I agree. I'm going to rework it like so:
Free users:
high = 128 mp3
Pandora One users:
high = 192 mp3
Everyone:
med = 64 aacplus
low = 32 aacplus

That way we don't have to deal with the adts stupidness...

JasonLG1979:
There @TingPing all fixed up so the changes are only in Pandora.py. No workarounds needed.

PR #287, opened 2016-04-06
Change Audio quality on the fly.

JasonLG1979:
When the audio quality is changed in the preferences(and apply is clicked)
this will load the url for the chosen audio quality, seek to where the
song left off and continue playing.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -115,10 +115,8 @@ class PlayerStatus:
     self.reset()
 
   def reset(self):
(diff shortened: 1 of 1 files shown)


JasonLG1979:
I'm going to close this. After testing for some reason it causes Pithos to use more RAM. Maybe something to do with Gstreamer not clearing buffers, idk? In any event this isn't really a "Must have feature." There are other more important things to be done.

PR #291, opened 2016-04-10
Add "Create Station" to the treeview song menu

JasonLG1979:
Adds the ability to create a new station based on the song or artist
in the treeview song menu.


JasonLG1979:
I'm going to close this also. It may be revisited in the future borrowing from the logic that checks for duplicate stations before adding them in the station search. There's just no point in getting and sorting through the station seeds if no meaningful comparisons can be made between the seeds and song the user is right clicking on. A better approach is to just always present the option and if a duplicate station is returned show a warning dialog like in the station search.

PR #302, opened 2016-07-03
Make use of the stream-start gstreamer message and clean up unneeded ones.

JasonLG1979:
The songs duration can be got at and we can check for ads at the posting of the stream-start message. This allows us to remove async-done and duration-change. We can also git rid of the tag message(that spams the heck out of debug mode). The only thing we use it for is getting the bitrate which Pandora already gives us.

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -393,7 +393,6 @@ class Song:
                     'audioUrl': d['additionalAudioUrl'][0],
                 }
 
-        self.bitrate = None
         self.is_ad = None  # None = we haven't checked, otherwise True/False
         self.tired=False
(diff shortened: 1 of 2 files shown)


TingPing:
Curious where you found docs on this? All I found was:

> Stream Start
>
>WRITEME

JasonLG1979:
https://lazka.github.io/pgi-docs/#Gst-1.0/flags.html#Gst.MessageType.STREAM_START

>Message indicating start of a new stream. Useful e.g. when using playbin in gapless playback mode, to get notified when the next title actually starts playing (which will be some time after the URI for the next title has been set).

I was playing with 'about-to-finish' and came across it.

PR #314, opened 2016-07-15
make: install plugins dbus_util

Stebalien:
There may be a better way to do this but I'd like to minimize my exposure to
autotools.


TingPing:
Thanks, silly mistake...

PR #320, opened 2016-07-27
MPRIS Fixup

JasonLG1979:
Fixes: TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'

For some reason self.window.query_duration() passes the None check and then shortly after returns None. This should fix that by caching the 1st call and reusing the value instead of calling it twice.

### DIFF
--- a/pithos/plugins/_mpris.py
+++ b/pithos/plugins/_mpris.py
@@ -147,8 +147,9 @@ class PithosMprisService(DBusServiceObject):
     def _duration(self):
         # use the duration provided by Pandora
         # if Gstreamer hasn't figured out the duration yet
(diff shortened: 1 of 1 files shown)


TingPing:
Welp, 1.2.1 :P

JasonLG1979:
Sorry, I just ran into it just now. I tried to get it in before the version bump.

TingPing:
No worries, sounds like that issues shouldn't happen often.

JasonLG1979:
It doesn't. It's the 1st time I noticed it.

JasonLG1979:
Merge maybe?

TingPing:
Calm down, I'll get around to it.

PR #331, opened 2016-07-30
Return a generator from get_playlist instead of a list

JasonLG1979:
get_playlist may as well return a generator because:
1. It's only iterated over once.
2. We don't use any list feature with it.
3. Generators are faster than lists.(probably a microscopic difference in this case)

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -328,12 +328,8 @@ class Station:
                         'stationToken': self.idToken,
                         'includeTrackLength': True,
                         'additionalAudioUrl': 'HTTP_32_AACPLUS,HTTP_128_MP3',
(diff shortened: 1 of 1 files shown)


TingPing:
I get what you are going for but I don't think this is a worth while change. The list is extremely small and each iteration doesn't do anything time consuming so there is no benefit. The downsides are that an API call returning a generator might be less obvious in this case especially on a getter. I know we are the only consumer of the API but again, no benefits.

JasonLG1979:
Right on. Understandable. I'm just going though stuff and trying to squeeze every bit I can out of it.

TingPing:
Just make sure you are using a profiler to see actual performance bottlenecks :P

PR #333, opened 2016-07-30
Use sorted in search

JasonLG1979:
sorted(list) is preferred over list.sort() since python 2.4.

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -278,9 +278,8 @@ class Pandora:
 
         l =  [SearchResult('artist', i) for i in results['artists']]
         l += [SearchResult('song',   i) for i in results['songs']]
-        l.sort(key=lambda i: i.score, reverse=True)
 
-        return l
+        return sorted(l, key=lambda i: i.score, reverse=True)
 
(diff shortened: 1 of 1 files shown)


TingPing:
> sorted(list) is preferred over list.sort() since python 2.4.

That is a blanket statement, The former allocates a new list in sorted order and the latter sorts the list in place. Do you have any benchmarks showing a performance increase?

JasonLG1979:
You're right. My bad. Further testing shows that a search can return a list 70ish long so I ran this to get a rough estimate of performance:

```
import timeit
setup = """
import random
random.seed(0)
lst = list(range(100))
random.shuffle(lst)
"""
stuff = timeit.repeat("lst.sort()", setup=setup, number = 10000)
print(stuff)
stuff2 = timeit.repeat("sorted(lst)", setup=setup, number = 10000)
print(stuff2)
```

That get me:

```
[0.02786459300000388, 0.024952423000058843, 0.023114420000069913]
[0.16670333900003698, 0.16613745099994048, 0.16613261700001658]
```

Which tells me that list.sort() is faster.

PR #334, opened 2016-08-03
Rewrite of notify.py

JasonLG1979:
1. Complete rewrite from scratch using Gio.DBusProxy.
2. Removed external (crufty) libnotify dependency.
3. Completely, 100% non-blocking/asynchronous method calls, including initialization.
4. Removed unused OSX support.(haven't been able to install Pithos in OSX in a while now.)
5. Made Notifications translatable where appropriate.

### DIFF
--- a/pithos/plugins/notify.py
+++ b/pithos/plugins/notify.py
@@ -1,6 +1,6 @@
-# -*- coding: utf-8; tab-width: 4; indent-tabs-mode: nil; -*-
-### BEGIN LICENSE
-# Copyright (C) 2010-2012 Kevin Mehall <km@kevinmehall.net>
+#
+# Copyright (C) 2016 Jason Gray <jasonlevigray3@gmail.com>
+#
 #This program is free software: you can redistribute it and/or modify it 
 #under the terms of the GNU General Public License version 3, as published 
 #by the Free Software Foundation.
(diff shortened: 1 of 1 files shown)


TingPing:
- I'd still prefer that this be put in its own subclass like my example.
- You have to edit po/POTFILES.in for translations
- Move unrelated changes like translations to its own commit
- Make a specific commit for like removing OSX support
- Unrelated header changes at the top
- Feel free to add your copyright line but don't remove Kevins, much of the code was moved not rewritten.
- Unrelated import change removing ()

JasonLG1979:
> I'd still prefer that this be put in its own subclass like my example.

Your example didn't work,lol!!! I could but then the main class would be like 10 lines long. I didn't see the point of separating them, or subclassing even for that matter?







> You have to edit po/POTFILES.in for translations
> Move unrelated changes like translations to its own commit
> Make a specific commit for like removing OSX support
> Unrelated header changes at the top

Ok.



> Feel free to add your copyright line but don't remove Kevins, much of the code was moved not rewritten.

Look again side by side besides the licence I kept may 10 lines of actual code and most of that is function names... But Ok.

> Unrelated import change removing ()

Didn't see the point of it so I got rid of it...

JasonLG1979:
Do you want each commit in it's own PR or do you want me to just separate the changes into commits in this PR? As a side note have you tested it out? It works like a charm on every DE I've tried it on.

TingPing:
This PR is fine. I've yet to test it.

JasonLG1979:
How about something like this for a wrapper @TingPing?:
https://gist.github.com/JasonLG1979/2e966d48ff0c25550f27a65f6f79a389

It mimics the libnotify bindings a lot so notify.py would not need to be changed very much at all.

TingPing:
Basically, and just throw it in the dbus_util dir.

JasonLG1979:
Right on.

TingPing:
protip: `GLib.Variant.new_...()` can be replaced by `GLib.Variant('...', )` using these type strings [GVariantType](https://developer.gnome.org/glib/stable/glib-GVariantType.html#GVariantType)

JasonLG1979:
> protip: GLib.Variant.new_...() can be replaced by GLib.Variant('...', ) using these type strings GVariantType

Thanks.

JasonLG1979:
For testing purposes here's notify.py with SimpleDBusNotifications in it.
https://gist.github.com/JasonLG1979/88d79eadd3a6a5d5b320bf21ed7d2308

JasonLG1979:
@TingPing: Fancy version(https://gist.github.com/JasonLG1979/2e966d48ff0c25550f27a65f6f79a389) or Simplified version (https://gist.github.com/JasonLG1979/7a35efc5ed94af448291995602d08ea5)?

TingPing:
Mix of both I guess.

- I think having the gsignals make a lot of sense.
- The properties.. might be overkill for us.
- I think having the default timeout and urgencies are fine.

Also eitherway I'm probably going to fiddle with the end result a bit anyway as to not annoy you with minor details :P

JasonLG1979:
Right on I figured you want something in between. The fancy version is way overkill. I just wanted to give you the option. I'll keep the signals and ditch the props. As far as tweeking it, I would expect no less ;p

JasonLG1979:
Another question: In the fancy version you can feed it a pixbuf as the 'icon' in the 'new' method. What it actually does is set the icon method arg to '' and gets the raw image data from the pixbuf and sets the 'image-data' hint with the raw data. To my knowledge it's the only blocking bit in the whole code:
```python
        def raw_data_from_pixbuf(icon):
            raw_data = GLib.Variant('(iiibiiay)',	
                                    (icon.get_width(),
                                     icon.get_height(),
                                     icon.get_rowstride(),
                                     icon.get_has_alpha(),
                                     icon.get_bits_per_sample(),
                                     icon.get_n_channels(),
                                     icon.get_pixels(),
                                    ),
            )

            self.set_property('image-data', raw_data)
            self.show()
```

If we just stick to 'icon = song.artUrl or 'audio-x-generic'' we don't really need that part and everything(AFAIK) goes back to non-blocking. I don't think we really need pixbuf support in it since we have an art url.

TingPing:
I don't think that would be blocking, the pixbuf is already in memory so those should be simple lookups?

JasonLG1979:
You'd think. I'm not sure what's at play, but there is a microscopic delay with the pixbuf where the url seems instant.

JasonLG1979:
If I had to guess the looks aren't blocking in them self but the bytes lookup returns a relatively large object that then has to be packed into a variant? The url string has got to be tiny compared to the raw_data variant?

TingPing:
Yea plausible it is just slow, as you said probably not needed for our usage anyway.

JasonLG1979:
Update still working on this. @TingPing what do you think about this:
https://github.com/JasonLG1979/possibly-useful-scraps/blob/master/SimpleDBusNotifications.py
https://github.com/JasonLG1979/possibly-useful-scraps/wiki/SimpleDBusNotifications

If that looks good I'll call it good and rework my pull request with step by step like you asked.

TingPing:
Curious why Signals is a separate class.

JasonLG1979:
> Curious why Signals is a separate class.

You can't 'self.emit' from the main class you'll get an error. Can't remember it off the top of my head. something about the method expecting a GLib.Variant? Maybe a consequence of subclassing Gio.DBusProxy?

JasonLG1979:
```bash
Traceback (most recent call last):
  File "/usr/lib64/python3.5/site-packages/gi/overrides/Gio.py", line 156, in __call__
    arg_variant = GLib.Variant(signature, tuple(args))
  File "/usr/lib64/python3.5/site-packages/gi/overrides/GLib.py", line 242, in __new__
    (v, rest_format, _) = creator._create(format_string, [value])
  File "/usr/lib64/python3.5/site-packages/gi/overrides/GLib.py", line 139, in _create
    raise NotImplementedError('cannot handle GVariant type ' + format)
NotImplementedError: cannot handle GVariant type 2
Traceback (most recent call last):
  File "/usr/lib64/python3.5/site-packages/gi/overrides/Gio.py", line 156, in __call__
    arg_variant = GLib.Variant(signature, tuple(args))
  File "/usr/lib64/python3.5/site-packages/gi/overrides/GLib.py", line 242, in __new__
    (v, rest_format, _) = creator._create(format_string, [value])
  File "/usr/lib64/python3.5/site-packages/gi/overrides/GLib.py", line 139, in _create
    raise NotImplementedError('cannot handle GVariant type ' + format)
NotImplementedError: cannot handle GVariant type 2
```

JasonLG1979:
Above if you try to emit a string.

```bash
Traceback (most recent call last):
  File "/usr/lib64/python3.5/site-packages/gi/overrides/Gio.py", line 152, in __call__
    raise TypeError('first argument must be the method signature string: %r' % signature)
TypeError: first argument must be the method signature string: 2
Traceback (most recent call last):
  File "/usr/lib64/python3.5/site-packages/gi/overrides/Gio.py", line 152, in __call__
    raise TypeError('first argument must be the method signature string: %r' % signature)
TypeError: first argument must be the method signature string: 2
```

If you try to emit an int.

TingPing:
Ah right, I overrode that in one of my classes:

```python
    def __getattr__(self, name):
        # pygobject ships an override that breaks our usage
        return object.__getattr__(self, name)
```

JasonLG1979:
> Ah right, I overrode that in one of my classes

I get that exception in my example not in Pithos.

JasonLG1979:
Edit: That fixes it. I'll change it and move the signals into the main class.

JasonLG1979:
Fixed. What do you think now @TingPing?

TingPing:
Looks good at a glance.

JasonLG1979:
> Looks good at a glance.

Cool. As you've said before I'm sure you'll want to tweak it a bit. When you do I'm going to split it out into it's own repo with your changes so that other apps can use it if they want. IMHO it's an improvement to the libnotify bindings.

JasonLG1979:
I'm still working on this. @TingPing I'm also adding a prefs dialog for the notify. What naming scheme for adding new setting values to io.github.Pithos.gschema.xml do you want to use I need to add a few bools.

TingPing:
Plugins have a `data` preference, see the lastfm plugin. It is a big ugly to store a lot of data in it but functional.

JasonLG1979:
You want me to store a bunch of bools as one crazy long string?

TingPing:
Could store it as JSON for example.

JasonLG1979:
That seems hacky and ugly. Why not just add a few new bools named something like bool-one-default-true and so forth?

JasonLG1979:
They would be reusable by other plugins ofc.

TingPing:
Eh, I think the json solution is actually the least hacky. GSettings only allows static definitions for settings so the most flexible way to expose that is by allowing them to store arbitrary text. Its contents are up to the plugin. We could add two bools but what happens when a plugin wants a third, or they want an integer and a float, etc?

JasonLG1979:
True. I guess I'll have to read up on JSON in python,lol!!!

TingPing:
Really easy to work with, just see `json.loads()` and `json.dumps()`

JasonLG1979:
Will do. Thanks.

JasonLG1979:
Yep, that was pretty easy. OK @TingPing I tried to match the layout of the main prefs window this look good to you as far as the sub-prefs dialog:
screenshot from 2016-09-16 18-29-24

The reset button isn't a must for a setting box that'll kave at most 3 setting boxes(only 2 are shown in the screenshot because they are generated dynamically based on the notification server caps and GNOME shell doesn't support icons in action buttons) I wanted to set kinda a style precedent of sorts for plugin prefs that may have more settings. What do you think?

Here's my WIP code:
 https://gist.github.com/JasonLG1979/8fe3327fc5ccb9ccc65ced232dac820a

JasonLG1979:
Here's a branch to test if you'd like:
https://github.com/JasonLG1979/pithos/tree/notify_rewrite_wip

JasonLG1979:
I'm closing this and submitting a new PR that I think will be more to your liking @TingPing.

PR #335, opened 2016-08-07
GBusService changes and discussion

infirit:
I have been poking at your ``DBusServiceObject`` class and saw it was missing some features. I am interested as I need something like this for a project trying to move over to GDBus. Please see this not as a PR to actually merge as you most likely have no need for the Python2 compat but to start a dialogue.

I need async callbacks and sender which is missing in the current implementation. I wanted to share some of the ideas here and perhaps collaborate a bit. We could as was done with dbus-python and here add more and more keywords to the decorator but  this lacks in flexibility and gets ugly with long list of keywords.

Instead I was thinking of having a wrapper object around the ``GDBusMethodInvocation`` which is then send to the method to do with as they please. The wrapper would have some convenience functions that handles the construction of the Variant return value, errors etc.

Also I am not sure we need use something like a GTask to handle the asynchronous? Also what if the function never return, should we then have a timeout of some sort?

Opinions, suggestions or tell me to go away :smile:

### DIFF
--- a/pithos/plugins/dbus_util/DBusServiceObject.py
+++ b/pithos/plugins/dbus_util/DBusServiceObject.py
@@ -40,12 +40,13 @@ class DBusArgInfo:
         self.signature = signature
 
 class DBusMethodInfo:
-    def __init__(self, name="", interface=None, in_args=[], out_args=[], sender="", annotations=[]):
+    def __init__(self, name="", interface=None, in_args=[], out_args=[], sender="", path="", annotations=[]):
         self.name = name
         self.interface = interface
         self.in_args = in_args
         self.out_args = out_args
         self.sender = sender
+        self.path = path
         self.annotations = annotations
 
     def generate_xml(self):
@@ -120,12 +121,14 @@ class DBusNodeInfo:
             node.append(interface.generate_xml())
         return node
 
-def _create_arginfo_list(func, signature, sender=None):
+def _create_arginfo_list(func, signature, sender=None, path=None):
     arg_names = inspect.getargspec(func).args
     signature_list = GLib.Variant.split_signature('(%s)' %signature) if signature else []
     arg_names.pop(0) # eat "self" argument
     if sender and sender in arg_names:
         arg_names.remove(sender)
+    if path and path in arg_names:
+        arg_names.remove(path)
 
     if len(signature_list) != len(arg_names):
         raise TypeError('Specified signature %s for method %s does not match length of arguments'
@@ -136,7 +139,7 @@ def _create_arginfo_list(func, signature, sender=None):
         args.append(DBusArgInfo(name=arg_name, signature=arg_signature))
     return args
 
-def dbus_method(interface, in_signature=None, out_signature=None, sender=None):
+def dbus_method(interface, in_signature=None, out_signature=None, sender=None, path=None):
     def decorator(func):
         in_args = _create_arginfo_list(func, in_signature, sender)
         out_args = [DBusArgInfo(name='return', signature=out_signature),] if out_signature else []
@@ -144,7 +147,8 @@ def dbus_method(interface, in_signature=None, out_signature=None, sender=None):
                                          interface=interface,
                                          in_args=in_args,
                                          out_args=out_args,
-                                         sender=sender)
+                                         sender=sender,
+                                         path=path)
         return func
 
     return decorator
@@ -332,6 +336,8 @@ class DBusServiceObject(GObject.Object):
 
         if info.sender:
             kwargs[info.sender] = sender
+        if info.path:
+            kwargs[info.path] = object_path
 
         try:
             ret = method(*args, **kwargs)
--- a/pithos/plugins/dbus_util/DBusServiceObject.py
+++ b/pithos/plugins/dbus_util/DBusServiceObject.py
@@ -255,10 +255,11 @@ class DBusServiceObject(GObject.Object):
                 attr = getattr(self, id)
             try:
                 info = attr._dbus_info
+                interface = self.__dbus_info.interfaces.setdefault(
+                    info.interface, DBusInterfaceInfo(name=info.interface))
             except AttributeError:
                 continue
-            interface = self.__dbus_info.interfaces.setdefault(info.interface,
-                                                              DBusInterfaceInfo(name=info.interface))
+
             if isinstance(info, DBusMethodInfo):
                 interface.methods.append(info)
             elif isinstance(info, DBusPropertyInfo):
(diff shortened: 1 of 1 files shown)


TingPing:
Which part of it is blocking and needs to be async?

JasonLG1979:
I think he's referring to:
https://github.com/pithos/pithos/blob/master/pithos/plugins/dbus_util/DBusServiceObject.py#L313-L333
Specifically:
https://github.com/pithos/pithos/blob/master/pithos/plugins/dbus_util/DBusServiceObject.py#L326-L333
If I'm reading that right(I may very well not be,lol!!!) that's basically a handler right? So what ever method your calling you're blocking waiting for a return value?

TingPing:
That is certainly running on the main thread but the docs say:

> Instances of the Gio.DBusMethodInvocation class are used when handling D-Bus method calls. It provides a way to asynchronously return results and errors.

And digging into the source this is correct and the returns use [Gio.DBusConnection.send_message](http://lazka.github.io/pgi-docs/index.html#Gio-2.0/classes/DBusConnection.html#Gio.DBusConnection.send_message) which is async.

JasonLG1979:
So you're not actually calling anything in that actual method? You're just processing stuff that has already completed?

infirit:
>Which part of it is blocking and needs to be async?

>And digging into the source this is correct and the returns use Gio.DBusConnection.send_message which is async.

Ah, misunderstood that it was already async, good :+1: 

That does leave open how to pass on the sender and possibly some others like object path. Doing keywords like I did for sender?

TingPing:
@infirit Seems reasonable.

infirit:
Actually it does block, take for example [this](https://gist.github.com/infirit/12c2a84e09c6552c1e40e7c3473867c6) dbus-python test service that shows what the ``async_callbacks`` are used for. Even though the dbus method calls are async the calling of the method by the implementation is not.

Method ``long_running`` will block in the GDBus ``DBusServiceObject`` until seconds are passed and potentially timeout if seconds is large enough.
Method ``long_running_async_ok`` does not block and does not timeout, the error version return an error.

I have not looked at dbus-python how this is done but this needs to be fixed as in pithos this will cause problems at some point when a dbus method call takes longer than the proxy timeout. We should either make it a default handling so no more ``async_callbacks`` keyword or add the keyword.

JasonLG1979:
How did you test this in Pithos?

infirit:
>How did you test this in Pithos?

I didn't but also don't have to. Even if you use a callback on the client proxy when the proxy reaches it's timeout you will fail in the callback. Granted most calls in Pithos will likely never reach the timeout (24 seconds is the default I think) but there are plenty of projects that do like blueman.

TingPing:
> Try it for yourself in https://github.com/pithos/pithos/blob/master/pithos/plugins/_mpris.py#L284 and add a time.sleep(30) and you will see it error.

So your point is that the method is blocking the dbus client? Because the fact the method runs on our main thread is on purpose and what we want.

infirit:
>So your point is that the method is blocking the dbus client?

Yes, and in my test example there is no reason to. This is the problem ``async_callbacks`` is solving in dbus-python.

>Because the fact the method runs on our main thread is on purpose and what we want.

Define "we", there are lots of projects who would object to a method call timing out because it runs a long running method that is not expected to return anything __ever__. Unless you are not interested in this going upstream this is going to be a roadblock.

Simon said (no pun intended):
>Methods that return asynchronously should be first-class citizens.
  In dbus-python, they're somewhat awkward (and in dbus-glib, they
  needed special annotations).

TingPing:
Yea it is totally reasonable to not wait on methods with no return. Would that even be possible with a return?

infirit:
>Would that even be possible with a return?

I think dbus-python does if I read the code properly (see below), I'll add a method to my test to be sure.

The magic I think is in [``_method_lookup``] (https://cgit.freedesktop.org/dbus/dbus-python/tree/dbus/service.py#n185), ``candidate_class`` is what eventually is [called](https://cgit.freedesktop.org/dbus/dbus-python/tree/dbus/service.py#n706), but I am struggling with understanding it :confused:. When there are callbacks it just returns with no check on the signature.

TingPing:
Oh and another question is returning errors. If we shortcut a method with no return then errors will never happen.

JasonLG1979:
The error would be returned in the async callback.

JasonLG1979:
In the stuff that I've been messing with I set the async callback for the close method to None because It doesn't return anything (or at least anything I care about.)

JasonLG1979:
Oops I misunderstood. if the method blocks or not should be handled by the app using the interface. Ofc if you put a sleep in the main thread the app is going to freeze up? I don't see how that's not expected? The only thing I can see is maybe handling IF the call is async somehow. Maybe a arg in the decorator that tells it not to expect a return value?

JasonLG1979:
>Yea it is totally reasonable to not wait on methods with no return. Would that even be possible with a return?

Just return None after the async call and do a if ret is None?

JasonLG1979:
@infirit if the method in question needs to be async here's a decorator that can be used inside the method to make it async.
https://gist.github.com/JasonLG1979/e3ab511a60b64f5cdaa0a643d3f3540f

JasonLG1979:
Async decorator usage example:
https://github.com/JasonLG1979/possibly-useful-scraps/blob/master/README.md#gobjectasync

infirit:
Thanks jason, although I don't think we need to go that far. Appreciate the feedback though.

I pushed a couple of commits to this PR and dropped the python compatibility one as it makes little sense for pithos.

The first is a bug fix for objects that will __always__ return an attribute.

The following two add the sender and the path. There are plenty more in [dbus-python](https://cgit.freedesktop.org/dbus/dbus-python/tree/dbus/service.py#n671) but I am not sure which others would make sense to add. For now just sender and object path.

The last is a WIP patch to handle functions that return asynchronously. There are two methods on the new DBusMethodCallback object, one for each one provided by dbus-pyhton. Feedback on this one welcome.

infirit:
Question on the comment in [connection setter](https://github.com/pithos/pithos/blob/master/pithos/plugins/dbus_util/DBusServiceObject.py#L290). The comment suggest there is a way to add and remove properties, signals and methods after initial export, but I see no "legal" way to do it, old comment or something to be implemented?

ps: Background, I have plugins that can be loaded and unloaded at runtime, some of them export dbus methods so it would be nice if there was standardized way to "refresh" :smile:.

TingPing:
>  The comment suggest there is a way to add and remove properties, signals and methods after initial export,

No it is about changing connections on the same object. Probably not a very common usage.

infirit:
Ah, yes, that makes more sense, thx.

PR #350, opened 2016-10-19
Add an "end-of-stream" signal

jameswmills:
Emit the "end-of-stream" signal when the "eos" message is received
from the gstreamer bus.  This change will allow plugins to perform
synchronous actions when this message is received.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -105,6 +105,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         "user-changed-play-state": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_BOOLEAN,)),
         "metadata-changed": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "buffering-finished": (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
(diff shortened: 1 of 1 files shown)


JasonLG1979:
Unless you have a plugin that would use the signal I don't see the point of adding it yet.

jameswmills:
@JasonLG1979, thanks for taking a look at this.  My goal here was to introduce the availability of a synchronous "handle" for this asynchronous event to facilitate the creation of plugins that might want to do things between the end of one song and the beginning of another.  If we can get this introduced, a potential plugin developer won't have to wonder "What if I do all this work and they don't accept the change to the core code that enables this?" :)

JasonLG1979:
I'm not the final word on it but generally unused code is considered dead code. From my experience @tingping has no problem adding signals that are actually being used though.

JasonLG1979:
As far as the what if. That's the risk you take. For every 10 half baked PR's I put in 1 might actually get merged. That's just how it is. Your basically guilty until proven innocent. It's @tingpings job to say no most of the time.

TingPing:
> If we can get this introduced, a potential plugin developer won't have to wonder "What if I do all this work and they don't accept the change to the core code that enables this?" :)

So as of right now Pithos does not even support loading plugins outside of its tree so any potential developers will be able to do this themselves and lets be honest here we don't exactly have a large amount of people writing plugins and we never will. The plugin API as-is is very raw and I don't like that but this PR does nothing to change this. As mentioned if you submit a plugin that uses it that is fine.

Issue #351, opened 2016-10-19
Support Pandora Plus and it's new replay song feature.


ShapeShifter499:
Pandora One has changed to Pandora Plus which includes a feature that allows you to replay a song as much as you want. Is this support planned for Pithos?

TingPing:
Well if it is a new API it needs to be figured out and documented and figure out what limitations they impose, etc.

JasonLG1979:
To sum it up the new features are:
Unlimited skips
Unlimited replays
Offline listening

As far as I can tell we get unlimited skips for free, no code change needed. I just skipped though about 20 - 30 playlists and I didn't get a API_ERROR_PLAYLIST_EXCEEDED error.

I'll check to see how the unlimited replays works. Best case the playlists just don't expire and we can just reload the song url as is.

ShapeShifter499:
Yea I don't really care about offline, if I wanted I could just load the app on my Android Phone.  I'm much more interested in being able to replay songs. Saves me the hassle of going to Youtube or other sources if I really wanted to hear the song again.

JasonLG1979:
It would be a neat feature. We just have to figure out how it works on Pandora's side. I'm in the process of trying to figure out how it works on Pandora's side right now.

JasonLG1979:
The mobile apps remember the last few songs played from every station. I'm not really interested in doing that, it could get complicated real fast. If/when I implement this it will more than likely just allow you to replay songs of your current playlist, and not remember anything. Basically it'll be just like it is now except you'll be able to double click and replay songs you've already listened to in addition to songs you haven't yet.

ShapeShifter499:
@JasonLG1979 If the song url never expires, maybe there can be a bookmark feature that would allow playback in the future? or is there legal matters that would get in the way of that?

TingPing:
The only potential legal issue I want to avoid is storing the song files on disk.

JasonLG1979:
It would be a PITA. It would involve storing all the metadata or figuring out how to re-retrieve it.

JasonLG1979:
1 step at a time. 1st we have to see how it works.

JasonLG1979:
It would be trivial to add a repeat and play next option to the right click menu though.

JasonLG1979:
@TingPing getting whether the user is a Plus user or not is just a matter of adding 'returnIsSubscriber': True to the auth.userLogin call in pandora.connect :
```python
[...]

PR #352, opened 2016-10-21
Changed PLAYLIST_VALIDITY_TIME in pandora.py to 1 hour.

JasonLG1979:
While doing some testing for #351 I have determined that a playlist is only good for about an hour.
This also changes the is_still_valid method in pandora.py to account for the length of the song to determine if the song is still valid.
A song is considered valid if there is enough time for the song to play in it's entirety before the playlist has expired.


TingPing:
Would the track length really be a factor? It fails if it expires mid-song?

JasonLG1979:
This will keep us from getting Gstreamer errors because the song will be considered invalid if there is not enough time for it to play before the playlist has expired. Without considering the track length. If the playlist was valid for 2 more mins and you started a 3 min song the song audio url would expire in the middle of the song and gstreamer would throw an error.

PR #353, opened 2016-10-21
Remove redundant user-changed-play-state signal

JasonLG1979:
user-changed-play-state is redundant because it is always followed by the play-state-changed which serves the same purpose.


TingPing:
In theory they do serve different purposes, one indicates it was the player who made the state change. The question is whether this matters for that plugin at all.

JasonLG1979:
In practice  it is always followed by the play-state-changed signal no matter what. And it's only used by 1 plugin ATM.

JasonLG1979:
The main reason for this PR is clean up for some some logic I'm working on that would check the validity of a song when going from play to pause to more gracefully indicate that a song is no longer valid if it has become invalid while it was paused. Mostly useful in conjunction with screensaver pause plugin. So when the user unlocked the screen Pithos would skip to a valid song without throwing a bunch of Gstreamer errors.

JasonLG1979:
Yep, my bad. Never mind it does actually matter for that plugin.

PR #354, opened 2016-10-22
Check the validity of a song when going from pause to play.

JasonLG1979:
Check the validity of a song when going from pause to play to more gracefully indicate that a song is no longer valid if it has become invalid while it was paused. Mostly useful in conjunction with the screensaver pause plugin or after other long periods in which Pithos might be left in a paused state. So when the user unlocked the screen Pithos would skip to a valid song without throwing a bunch of Gstreamer errors.

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -324,16 +324,17 @@ class Station:
 
     def get_playlist(self):
         logging.info("pandora: Get Playlist")
+        # Set the playlist time to the time we requested a playlist.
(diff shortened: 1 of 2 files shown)


JasonLG1979:
I think I'm done messing with this. @TingPing What do you think?

TingPing:
I'll have to test it out when I get a chance.

JasonLG1979:
It worked pretty good for me. I tested it by playing a song for a while and  I then left Pithos paused for an hour. When I hit play it skipped the invalid songs and grabbed a new playlist.

PR #357, opened 2016-10-23
Fix song.duration being set to None.

JasonLG1979:
Very rarely query_duration returns None when called in on_gst_stream_start resulting in song.duration being set to None which causes check_if_song_is_ad to fail and no time info to be displayed in the UI. This fixes that by falling back to song.trackLength if query_duration returns None.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -762,7 +762,13 @@ class PithosWindow(Gtk.ApplicationWindow):
         return duration
 
     def on_gst_stream_start(self, bus, message):
-        self.current_song.duration = self.query_duration()
+        # Very rarely query_duration returns None when called in on_gst_stream_start
+        # causing song.duration to be set to None.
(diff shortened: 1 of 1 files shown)


JasonLG1979:
@TingPing another tiny bug fix. I'm not trying to push, I know you got other stuff going on and all but it would be awesome if you could check out my last few PR's and merge or close them :)  thanks in advance.

pithos/pithos.py
```diff
@@ -762,7 +762,7 @@ def query_duration(self):
         return duration
 
     def on_gst_stream_start(self, bus, message):
-        self.current_song.duration = self.query_duration()
+        self.current_song.duration = self.query_duration() or self.current_song.trackLength * 1000000000
```
TingPing: Use `Gst.SECOND`

JasonLG1979:
Better?

PR #359, opened 2016-11-01
Find Missing Covers Plugin.

JasonLG1979:
This Plugin attempts to find missing album cover art from Last.fm. If Album art can not be found an artist image is used instead if found.

Features:
Does NOT depend on pylast.
A UI for viewing, editing and optionally blacklisting artist and album corrections.
The corrections are searchable, sort-able and editable.
Art urls are cached and when reused are verified to still be valid.
Context menu to copy contents of cell and search Last.fm.

This Plugin would be a nice addition to Pithos but I admit it is HUGE. It would be nice if we could implement out of tree plugins.


TingPing:
Sorry but as I mentioned elsewhere I just have no interest in this. As you say it is a lot of code for a feature that I see little value in (I almost never see missing artwork). I barely review other PRs as is so I'll just close this.

JasonLG1979:
Fair enough. If/when Pithos allows loading external plugins this would be a good plugin.

PR #362, opened 2016-11-05
Improved station search dialog

JasonLG1979:
Make better use of GtkTemplate
Moved search entry to headerbar
Added placeholder text to search entry
model is no longer populated if the search entry is empty

### DIFF
--- a/pithos/SearchDialog.py
+++ b/pithos/SearchDialog.py
@@ -15,52 +15,59 @@
 ### END LICENSE
 
 import html
-from gi.repository import GObject, Gtk
+from gi.repository import GObject, Gtk, Gdk
 
 from .gi_composites import GtkTemplate
 
 @GtkTemplate(ui='/io/github/Pithos/ui/SearchDialog.ui')
 class SearchDialog(Gtk.Dialog):
     __gtype_name__ = "SearchDialog"
+    __gsignals__ = {
+        "add-station": (GObject.SignalFlags.RUN_FIRST, None, (str,)),
+    }
 
     entry = GtkTemplate.Child()
-    treeview = GtkTemplate.Child()
+    search_results_model = GtkTemplate.Child()
 
-    def __init__(self, *args, **kwargs):
-        self.worker_run = kwargs["worker"]
-        del kwargs["worker"]
-
-        super().__init__(*args, use_header_bar=1, **kwargs)
+    def __init__(self, worker_run):
+        super().__init__()
         self.init_template()
-
-        self.model = Gtk.ListStore(GObject.TYPE_PYOBJECT, str)
-        self.treeview.set_model(self.model)
-
-        self.result = None
+        self.worker_run = worker_run
+        self.query = ''
 
     @GtkTemplate.Callback
-    def search_clicked(self, widget):
-        self.search(self.entry.get_text())
+    def on_delete_event(self, *ignore):
+        self.query = ''
+        self.hide()
+        # Reset everything.
+        text_in_search_entry = self.entry.get_text()
+        if text_in_search_entry:
+            self.entry.set_text('')
+        self.search_results_model.clear()
+        return True
 
     @GtkTemplate.Callback
-    def get_selected(self):
-        sel = self.treeview.get_selection().get_selected()
-        if sel[1]:
-            return self.treeview.get_model().get_value(sel[1], 0)
-
-    def search(self, query):
-        if not query: return
+    def on_entry_search_changed(self, entry, model):
+        self.query = entry.get_text()
+        if not self.query: return
         def callback(results):
-            self.model.clear()
+            model.clear()
+            # If there is no current query(the searchbox is empty)
+            # we don't want to populate the model with irrelevant results.
+            if not self.query: return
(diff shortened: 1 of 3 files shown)


TingPing:
> Moved search entry to headerbar

Seems a bit crowded to me.

JasonLG1979:
> Seems a bit crowded to me.

What do you mean?

JasonLG1979:
screenshot from 2016-11-05 17-36-28
screenshot from 2016-11-05 17-32-27

IMHO it makes better use of screen space and makes it simpler(not that it was very complex to begin with ;P)

TingPing:
I think it looks worse, also continuing the accessibility talk this is a downgrade (Other than the placeholder text), as there is no discoverable way to close the window or add the station as the buttons previously provided.

JasonLG1979:
I can add a widow title(it won't be visible but the window will have a title) . As far as buttons widow close button serves the same function as "Cancel" and "Add" isn't needed. When the treeview row is clicked the station is added and the dialog is closed(hidden).

TingPing:
>  As far as buttons, the widow's close button serves the same function as "Cancel" and "Add" isn't needed. 

Well currently *Add* is actually impossible without a mouse. We need to bind activate or something so the keyboard can add. *Cancel* is only available via alt+f4 which is less than ideal and again not discoverable.

pithos/SearchDialog.py
```diff
@@ -15,52 +15,59 @@
...
-    treeview = GtkTemplate.Child()
+    search_results_model = GtkTemplate.Child()
 
-    def __init__(self, *args, **kwargs):
-        self.worker_run = kwargs["worker"]
-        del kwargs["worker"]
-
-        super().__init__(*args, use_header_bar=1, **kwargs)
+    def __init__(self, worker_run):
+        super().__init__()
```
TingPing: Always pass `**kwargs` to init, pygobject uses it for setting constructor properties.

pithos/StationsDialog.py
```diff
@@ -38,7 +38,9 @@ def __init__(self, pithos, *args, **kwargs):
         self.model = pithos.stations_model
         self.worker_run = pithos.worker_run
         self.quickmix_changed = False
-        self.searchDialog = None
+        self.searchDialog = SearchDialog.SearchDialog(self.worker_run)
+        self.searchDialog.set_transient_for(self)
```
TingPing: Once the `__init__()` func is fixed as mentioned above. `SearchDialog(self.worker_run, transient_for=self)`

JasonLG1979: OK

JasonLG1979:
> Well currently Add is actually impossible without a mouse. We need to bind activate or something so the keyboard can add. Cancel is only available via alt+f4 which is less than ideal and again not discoverable.

idk about "Add" but "Cancel" could just be removed and replaced with the window's close button. To me it's understood that closing a window means take no action which in this case would be the same as "Cancel".

pithos/SearchDialog.py
```diff
@@ -15,52 +15,59 @@
...
-        self.result = None
+        self.worker_run = worker_run
+        self.query = ''
 
     @GtkTemplate.Callback
-    def search_clicked(self, widget):
-        self.search(self.entry.get_text())
+    def on_delete_event(self, *ignore):
+        self.query = ''
+        self.hide()
```
TingPing: http://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/Widget.html#Gtk.Widget.hide_on_delete

pithos/SearchDialog.py
```diff
@@ -15,52 +15,59 @@
...
-            self.treeview.show()
-        self.worker_run('search', (query,), callback, "Searching...")
+                model.append((i.musicId, mk))
+        self.worker_run('search', (self.query,), callback, "Searching...")
 
-    def cursor_changed(self, *ignore):
-        self.result = self.get_selected()
-        self.set_response_sensitive(Gtk.ResponseType.OK, not not self.result)
+    @GtkTemplate.Callback
+    def on_treeview_button_press_event(self, treeview, event):
```
TingPing: http://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/TreeView.html#Gtk.TreeView.signals.row_activated

JasonLG1979:
> I'd rather not keep repeating myself but I think for this specific change the previous design was better.

OK, what about just moving the search entry up into the headerbar to make better use of screen space?

TingPing:
Why close this?

JasonLG1979:
Because from now on all UI changes have to take accessibility into account and since we don't know what that actually entails I'm not going to waste my time on it.

TingPing:
Well you did other changes that don't involve UI changes.

Also the guidelines to be accessible aren't that hard:

- Make everything navigable via keyboard
- If something isn't text, like icons, give it a description

My problems with this UI change was clear:

- You removed textual descriptions
- It is not navigable by keyboard

JasonLG1979:
The only agreeable change was the placeholder text. I can put in a separate PR just for that.

PR #364, opened 2016-11-06
Add placeholder text to new station search dialog entry

JasonLG1979:


data/ui/SearchDialog.ui
```diff
@@ -32,7 +32,7 @@
               <object class="GtkSearchEntry" id="entry">
                 <property name="visible">1</property>
                 <property name="can_focus">1</property>
-                <property name="invisible_char">•</property>
+                <property name="placeholder_text" translatable="yes">Search for an Artist or Song...</property>
```
TingPing: Use `…` rather than `...`

JasonLG1979: Done

PR #371, opened 2016-11-15
Window size and maximization is now preserved between runs of Pithos

ChaiTRex:
While Pithos remembers the position of the upper-left corner of the Pithos window in between runs of Pithos, it doesn't remember the size of the window, which makes it where I have to manually resize it every time I run it.

This pull request causes Pithos to remember the size of the window between runs.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -1117,6 +1117,12 @@ class PithosWindow(Gtk.ApplicationWindow):
         x, y = self.settings['win-pos']
         if not x is None and not y is None:
             self.move(int(x), int(y))
+        width, height = self.settings['win-size']
+        if not width is None and not height is None:
+            self.resize(int(width), int(height))
+        maximized = self.settings['win-max']
+        if maximized:
+            self.maximize()
 
     def bring_to_top(self, *ignore):
         self.set_initial_pos()
@@ -1127,6 +1133,18 @@ class PithosWindow(Gtk.ApplicationWindow):
     def on_configure_event(self, widget, event, data=None):
         self.settings.set_value('win-pos', GLib.Variant('(ii)', (event.x, event.y)))
         return False
+
+    @GtkTemplate.Callback
+    def on_window_resize(self, container, data=None):
+        maximized = container.get_window().get_state() & Gdk.WindowState.MAXIMIZED == Gdk.WindowState.MAXIMIZED
+        if not maximized:
+            width, height = container.get_size()
+            self.settings.set_value('win-size', GLib.Variant('(ii)', (width, height)))
+
+    @GtkTemplate.Callback
+    def on_window_state_event(self, widget, event, data=None):
+        maximized = widget.get_window().get_state() & Gdk.WindowState.MAXIMIZED == Gdk.WindowState.MAXIMIZED
+        self.settings.set_value("win-max", GLib.Variant('b', maximized))
 
     def quit(self, widget=None, data=None):
         """quit - signal handler for closing the PithosWindow"""
--- a/data/io.github.Pithos.gschema.xml
+++ b/data/io.github.Pithos.gschema.xml
@@ -47,6 +47,16 @@
       <summary>Position of window</summary>
     </key>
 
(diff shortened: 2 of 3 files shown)


JasonLG1979:
You should probably put that in the UI file and use the @GtkTemplate.

pithos/pithos.py
```diff
@@ -1128,6 +1133,10 @@ def on_configure_event(self, widget, event, data=None):
         self.settings.set_value('win-pos', GLib.Variant('(ii)', (event.x, event.y)))
         return False
```
JasonLG1979: @GtkTemplate.Callback

pithos/pithos.py
```diff
@@ -127,6 +127,8 @@ def __init__(self, app, test_mode):
         super().__init__(application=app)
         self.init_template()
 
+        self.connect('check-resize', self.on_window_resize)
```
JasonLG1979: Get rid of this.

pithos/pithos.py
JasonLG1979: And connect the signal handler in https://github.com/pithos/pithos/blob/master/data/ui/PithosWindow.ui

JasonLG1979: Under
https://github.com/ChaiTRex/pithos/blob/master/data/ui/PithosWindow.ui#L16

Put:
```xml
    <signal name="check-resize" handler="on_window_resize" swapped="no"/>
```

TingPing: This should be in the `configure-event` callback just like the position is.

JasonLG1979: Easy enough I was just going off what signal they were using. That simplifies things.

ChaiTRex:
OK, I've:

0. Added `@GtkTemplate.callback`
0. Moved the signal handler connection to `PithosWindow.ui`
0. Added maximization support, since maximizing and then quitting was screwing things up (window would forget its unmaximized size and be as large as it's possible to be without maximization)

I'm new to Python and GTK, so I'm not sure how to combine everything into one callback like @TingPing suggested.

JasonLG1979:
Sorry scrap what a said. @TingPing wants everything that saves settings in the configure-event handler. Also you wouldn't want to try to resize the window every time you brought it to the front. In my experience every window manager I've ever used has no problem remembering the size of a running window. Resizing the window would be a one shot at start up deal.

What you're looking for is something like this:

https://github.com/JasonLG1979/pithos/commit/b335b0051182064b09f6ce7ebd4fdf9018ee7f03

TingPing:
Slightly off topic but should Pithos even allow being maxmized? The UI was clearly not designed for it.

JasonLG1979:
@TingPing I agree 100% but I don't think there's a way to disallow maximizing that all window managers follow. I've messed with it in Tk and depending on the WM I got inconsistent behavior. Could have just been Tk it's a pretty basic toolkit?

TingPing:
@JasonLG1979 Looks like `gtk_window_set_resizable()` was what I was thinking of which is not quite what we want.

JasonLG1979:
@TingPing It would also be really easy to remember if the window was minimized when Pithos closed. Just make the 'min-max' a string and set the value to 'min', max' or '' or make min-max an int and use 0 - 2 to represent min, normal, max. But I'm not sure about starting Pithos minimized. It could be confusing and like you said the Pithos isn't really made to be maximized. I'm not even sure I like remembering the state at all. I still like the idea of keeping track of if Pithos is maximized for the purpose of position and size so it can properly be ignored though.

TingPing:
> But I'm not sure about starting Pithos minimized. It could be confusing

My thoughts exactly.

JasonLG1979:
It might be a good pref option but not something you'd want to do by default. I don't know about you but most of the time Pithos isn't focused.

JasonLG1979:
@TingPing your thoughts on even remembering state?

JasonLG1979:
This is almost a year old now with no activity and it's questionable if it's even very useful. I'm going to go a head and close this. @TingPing or @ChaiTRex you guys can re-open it if you'd like?

PR #383, opened 2016-12-30
Gst changes

JasonLG1979:
Simplify buffering,  increase souphttpsrc timeout, enumerate Gst.State's.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -188,6 +188,11 @@ class PithosWindow(Gtk.ApplicationWindow):
         self._query_buffer = Gst.Query.new_buffering(Gst.Format.PERCENT)
 
         self.player = Gst.ElementFactory.make("playbin", "player")
+        # Render the audio stream | Use software volume
+        self.default_player_flags = (1 << 1 | 1 << 4)
+        # Render the audio stream | Use software volume | Attempt progressive download buffering
+        self.downloadbuffer_player_flags = (1 << 1 | 1 << 4 | 1 << 7)
+        self.player.props.flags = self.default_player_flags
 
         bus = self.player.get_bus()
         bus.add_signal_watch()
@@ -595,6 +600,15 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.emit('song-changed', self.current_song)
         self.emit('metadata-changed', self.current_song)
 
+    def set_player_flags(self):
+        current_player_flags = self.player.props.flags
+        if self.settings['downloadbuffering']:
+            desired_flags = self.downloadbuffer_player_flags
+        else:
+            desired_flags = self.default_player_flags
+        if current_player_flags != desired_flags:
+            self.player.props.flags = desired_flags
+
     @GtkTemplate.Callback
     def next_song(self, *ignore):
         if self.current_song_index is not None:
@@ -647,8 +661,11 @@ class PithosWindow(Gtk.ApplicationWindow):
             self.emit("song-ended", prev)
 
         self.destroy_ui_loop()
-        self._set_player_state(PseudoGst.STOPPED)
-        self.emit('play-state-changed', False)
+        if self._set_player_state(PseudoGst.STOPPED, change_gst_state=True):
+            # Probably best to only change the flags when the pipeline
+            # is in a NULL state.
+            self.set_player_flags()
+            self.emit('play-state-changed', False)
 
     @GtkTemplate.Callback
     def user_playpause(self, *ignore):
@@ -835,6 +852,8 @@ class PithosWindow(Gtk.ApplicationWindow):
                         submsg=_("The required codec failed to install. Either manually install it or try another quality setting."))
 
     def on_gst_element(self, bus, message):
+        if message.has_name('GstCacheDownloadComplete'):
+            logging.debug('Download buffering complete')
         if GstPbutils.is_missing_plugin_message(message):
             if GstPbutils.install_plugins_supported():
                 details = GstPbutils.missing_plugin_message_get_installer_detail(message)
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -28,6 +28,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from gettext import gettext as _
+from enum import Enum
 
 import gi
 gi.require_version('Gst', '1.0')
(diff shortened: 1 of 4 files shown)


pithos/pithos.py
```diff
@@ -86,14 +87,24 @@ def do_render(self, ctx, widget, background_area, cell_area, flags):
...
             ctx.paint()
 
-class PlayerStatus:
-  def __init__(self):
-    self.reset()
-
-  def reset(self):
-    self.began_buffering = None
-    self.buffer_percent = 0
+class PseudoGst(Enum):
```

pithos/pithos.py
```diff
@@ -184,11 +197,11 @@ def init_core(self):
         self.player.connect("notify::volume", self.on_gst_volume)
         self.player.connect("notify::source", self.on_gst_source)
 
-        self.player_status = PlayerStatus()
-
         self.stations_dlg = None
 
-        self.playing = None # None is a special "Waiting to play" state
+        self._current_state = PseudoGst.STOPPED
+        self._previous_state = PseudoGst.STOPPED
```
* [W291] trailing whitespace

pithos/pithos.py
```diff
@@ -570,30 +591,41 @@ def next_song(self, *ignore):
...
+            ret = self.player.set_state(target.State)
+            if ret != Gst.StateChangeReturn.FAILURE:
+                if self._current_state is not PseudoGst.BUFFERING:
+                    # We only want the previous non-buffering state.
+                    self._previous_state = self._current_state
+                self._current_state = target
+                return True
+            current_state = self.player.state_get_name(self._current_state)
+            target_state = self.player.state_get_name(target.State)
+            logging.warning('Error changing player state from: {} to: {}'.format(current_state, target_state))
```
* [W291] trailing whitespace

pithos/pithos.py
JasonLG1979: OK

pithos/pithos.py
```diff
@@ -862,30 +915,29 @@ def on_gst_volume(self, player, volumespec):
...
-            if song.position is not None and song.duration is not None:
-                pos_str = self.format_time(song.position)
-                msg.append("%s / %s" % (pos_str, song.duration_message))
-                if self.playing == False:
-                    msg.append("Paused")
-            if self.player_status.buffer_percent < 100:
-                msg.append("Buffering (%i%%)" % self.player_status.buffer_percent)
+            pos_str = self.format_time(self.query_position() or 0)
+            msg.append("%s / %s" % (pos_str, song.duration_message))
+            if self.playing == False:
```
* [E712] comparison to False should be 'if cond is False:' or 'if not cond:'

TingPing:
Not sure I like options for such technical details as buffering method

JasonLG1979:
When it works it works great. Less memory usage, vastly improved network resilience and such.

TingPing:
Was the aacplus codec the only time it didn't work? Maybe automatically enable it for the mp3 qualities?

JasonLG1979:
It really depends on the Gstreamer version and the system. You can have to distros with the same Gstreamer version and one will work and the other doesn't. That's why I wanted to make it a user option. There just doesn't seem to be any hard and fast rhyme or reason for us to decide if/when to enable or disable it. This way the user can check the box and if it works great if not uncheck the box.

JasonLG1979:
*I mean there's no way for us to programmatically tell if it's OK to enable it or not.

pithos/pithos.py
```diff
@@ -780,6 +852,13 @@ def on_gst_plugin_installed(self, result, userdata):
                         submsg=_("The required codec failed to install. Either manually install it or try another quality setting."))
 
     def on_gst_element(self, bus, message):
+        if message.has_name('GstCacheDownloadComplete'):
+            logging.debug('Download buffering complete')
+            # In case we miss a buffering message.
+            # (known to happen with download buffering)
+            # when download buffering has completed
+            # it's safe to assume we're done buffering.
```
* [W291] trailing whitespace

JasonLG1979:
@TingPing That last commit should solve the issue where Pithos just stalls in a buffering state sometimes with download buffering. Basically about 50% of the issues related buffering before we removed download buffering.

TingPing:
On a related note I am still interested in moving the player logic to its own class so it is possible to reason about this mess.

JasonLG1979:
I can give it a shot.

JasonLG1979:
I've been playing with custom piplines and in my experiments I moved the core player logic out of the main class. It's not that hard. How separate do you want it?

TingPing:
Basically `pithos.py` should have no knowledge of GStreamer at all, it should just have a generic player instance with the minimum of properties/functions like play/pause/volume/uri. You might need to rework some signals in the plugins to connect to the player but otherwise shouldn't be too painful.

JasonLG1979:
To me the main class should also be separated from the UI logic. The main class should just be basically a skeleton connecting and passing signals to and from the other classes.

TingPing:
So I've been working on this a lot recently: https://github.com/TingPing/transmission-remote-gnome

The high level overview of that is I have a Client class that does all the HTTP work and the Application class keeps an instance of that, it passes that to the Window class which hooks up the few signals from the client it cares about and just displays that info.

The only addition we have here is our Window would also have a player instance. Does that sound reasonable to you?

JasonLG1979:
I'll have to check it out but I think we're on the same page.

JasonLG1979:
Model–view–controller basically?

TingPing:
Well, to a degree. The window does to some degree act as both the View and Controller (as UI actions cause all state changes anyway) but the Model is very cleanly separate from the UI.

JasonLG1979:
The more modular the better.

JasonLG1979:
Do you want me to do the moving of the player logic in this branch?

TingPing:
This might be the best description of it: https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter

The Window class is basically the Presenter, the actual Widgets it contains is the View, the Client is the model.

> Do you want me to do the moving of the player logic in this branch?

I would prefer you migrate the current code-base flaws and all first. Then clean-ups after the fact should be more self-contained and reviewable.

JasonLG1979:
OK where does this branch sit then?

TingPing:
Probably going nowhere and will need rebasing if you plan on doing this.

JasonLG1979:
> Probably going nowhere and will need rebasing if you plan on doing this.

Well crap... I guess I'll close this then and work on separating the player logic...

PR #386, opened 2017-01-02
Improved buffering behavior

JasonLG1979:
This improves buffering behavior in a few ways.

1st, It changes how we deal with buffering signals. Since the buffering percent that is received in the buffering message is pretty much useless this uses a buffering query to get a bool representing if the pipeline is buffering or not. We then respond to that bool accordingly.

2nd, It Enumerates the player states to give us a BUFFERING state. We can then use those Enums to clearly differentiate between PAUSED and BUFFERING.

3rd, It makes the timeout longer in souphttpsrc to give the user time to reconnect in the case of a network disconnect.

And finally it makes Pithos behave as you would expect during buffering as far as user initiated state changes. As in the pipeline doesn't actually change state until buffering completes at which time it goes to the correct state, whatever that may be.

Another positive side effect of this is that it puts the end to the UI thrashing during buffering.


JasonLG1979:
@TingPing I know it hasn't been open long but have you had a chance to check this out?

TingPing:
Oh I made comments a week ago but I didn't realize how the github review stuff actually worked.

pithos/pithos.py
```diff
@@ -52,6 +53,27 @@
...
+
+class PseudoGst(Enum):
+    """Create aliases to Gst.State so that we can add our own BUFFERING Pseudo state"""
+    PLAYING = 1
+    PAUSED = 2
+    BUFFERING = 3
+    STOPPED = 4
+
+    @property
+    def State(self):
```
JasonLG1979: I can fix that no problem

pithos/pithos.py
```diff
@@ -215,6 +231,16 @@ def init_core(self):
             self.tempdir = None
             logging.warning('Failed to create a temporary directory')
 
+    @property
+    def playing(self):
+        # Recreate the old "playing" attribute as a property.
+        # Track self._buffer_recovery_state because that's the state
+        # we wish we were in.
+        if self._buffer_recovery_state is PseudoGst.PAUSED:
```
JasonLG1979: ...and that.

JasonLG1979: Wouldn't that be ```return self._buffer_recovery_state is not PseudoGst.PAUSED``` since it's an enum?

pithos/pithos.py
```diff
@@ -818,38 +866,36 @@ def check_if_song_is_ad(self):
...
-                self.player.set_state(Gst.State.PLAYING)
-            else:
-                logging.debug("Buffer recovery. User paused")
-            # Tell everyone to update their clocks after we're done buffering or
-            # in case it takes a while after the song-changed signal for actual playback to begin.
-            self.emit('buffering-finished', self.query_position() or 0)
-            self.player_status.began_buffering = None
-        self.player_status.buffer_percent = percent
-        self.update_song_row()
+            if self._current_state is PseudoGst.BUFFERING:
```
JasonLG1979: I guess I don't understand your question here. self._buffer_recovery_state can only be PseudoGst.STOPPED, PseudoGst.PLAYING or PseudoGst.PAUSED. That's why there's a couple elif's.

TingPing: you have:

```python
if:
else:
    if:
```

That can be simplified.

JasonLG1979: Something like this?

```python
        buffering = self.query_buffer()

        if buffering and self._current_state is not PseudoGst.BUFFERING:
            logging.debug("Buffer underrun")
            if self._set_player_state(PseudoGst.BUFFERING):
                logging.debug("Pausing pipeline")
        elif not buffering and self._current_state is PseudoGst.BUFFERING:
            logging.debug("Buffer overrun")
            ...
```

JasonLG1979: I changed it to the above. Let me know if that works for you.

PR #388, opened 2017-01-04
Style cleanup for most of the plugins.

JasonLG1979:
Silences most of the flake8 output for the plugins.


JasonLG1979:
@TingPing, I didn't touch DBusServiceObject.py. I figured that was your baby and You'd want to do the clean up.

PR #390, opened 2017-01-08
Stops Pithos from quitting when loggin and Pandora One is changed

JasonLG1979:
A bit of an edge case but I have 2 accounts one account is my main listening account, it is a Pandora Plus(or One) account the other is a free account. They both have the same password. If I change both the Pandora One checkbox and the loggin without changing the password Pithos quits with a "No email or no password set!" error. This stops that from happening and Pithos behaves as expected and logs into the other account.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -437,7 +437,7 @@ class PithosWindow(Gtk.ApplicationWindow):
 
 
         email = self.settings['email']
-        password = get_account_password(email)
+        password = get_account_password(email) or self.prefs_dlg.password_entry.get_text()
         if not email or not password:
             # You probably shouldn't be able to reach here
             # with no credentials set


TingPing:
>  If I change both the Pandora One checkbox and the loggin without changing the password Pithos quits with a "No email or no password set!" error. This stops that from happening and Pithos behaves as expected and logs into the other account.

That is the bug that should be solved then, and I thought I did.

JasonLG1979:
Nope. I'm not exactly sure why but you can change either the login or the Pandora One checkbox and everything works as expected but if you change both Pithos quits every time.

JasonLG1979:
The only thing I can think of is it that since both the pandora one and login "changed" signal are connected to pandora_reconnect changing both in one go causes a double hit to pandora_reconnect.

JasonLG1979:
I got nothing on this beyond what's there, but what's there is incomplete. It's just a PITA to make sure the prefs are applied correctly. If I had a say, I'd say merge the logging plugin and call it a new release.

PR #391, opened 2017-01-08
Move Pulseaudio props to application.py

JasonLG1979:
And set the name, version and icon_name props. The icon_name prop is needed so that a Pithos icon is shown in the GNOME sound settings.

### DIFF
--- a/pithos/application.py
+++ b/pithos/application.py
@@ -12,7 +12,7 @@
 # You should have received a copy of the GNU General Public License along
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-
+import os
 import sys
 import signal
 import logging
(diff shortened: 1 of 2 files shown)


TingPing:
I do prefer `os.environ`

JasonLG1979:
> I do prefer os.environ

Why when there's a GLib function that does the same thing and it removes the need to import os?

JasonLG1979:
I really don't care either way. I thought you'd prefer GLib.setenv. I can change it to os.environ if you'd like?

TingPing:
environ being a dictionary you can directly modify is cleaner than a function call. It isn't like `os` is an external module so there is no downside.

JasonLG1979:
There I changed it. Better?

Issue #385, opened 2017-01-01
feature request - ability to add a station by artist and/or song when right-clicking on a playlist entry


neoh4x0r:
Right now the only way to add a station would be to open the station manager, click the 'add station' + button and then search for an artist or song.

But I was hoping it wouldn't be, too, difficult to add this functionally to the playlist's context menu.

If a song is playing and you like the artist and/or song you could then right-click the entry and add
a new station based on either the artist or just the song (i'd like to see both options).

Thanks.

JasonLG1979:
I actually started work on that a few months ago. Never finished though.
https://github.com/pithos/pithos/pull/291

The reason being is how Pandora stations work. Basically stations start from seeds(artist, album or gene). Where you can have a station with multiple seeds you can not have 2 stations that contain the same seed. So you can have station 1 with seeds a, b, c and station 2 with seeds x, y, z you can't create a new station with seeds a, b, c, x, y, z. If you try it will silently fail. It won't throw any errors or anything it just won't work. You can get seed Id's from the stations and you can create stations based on either the song or artist with the trackToken but it's not clear how to compare seed Id's and trackTokens so that if you right click a song and you already have a station with that song or artist it will only show the appropriate menu options. Directly comparing song metadata to the seed description doesn't even work because it's inconsistent. I can create a station artist based station for "Some Random Artist" and the returning seed description string may or may not match "Some Random Artist".

neoh4x0r:
@JasonLG1979 I only say that because the official Pandora apps on my TV (as well as the web based app) can add a station based on both the artist and track.

I did this for an artist I was listening to in pithos through the webapp, and then synced the stations in pithos.
Basically by selecting 'Song info' from the context menu which launched my browser -- from there I could use Pandora's web interface to add artist as a station (and pithos's stations list was updated when I refreshed it).

Are you saying that this is not possible with Pandora or just with Pithos?

JasonLG1979:
We aren't an official client. We rely on the reverse engineering of others and our own tinkering to figure out the Pandora API. We don't have access to the real API docs like the the people who write official Pandora apps.

The best I can do at this point is always present the option to create a station based on the song and artist for every song and not check anything. Since it silently fails it doesn't hurt anything except a wasted network call. But then I'd get a bunch of bug reports about how people tried to add a station but it didn't show up.

JasonLG1979:
Don't get me wrong I think it's a great feature. There are a lot of feature that the official apps have that I'd love to add.

[...]

PR #395, opened 2017-01-13
Show dialog when trying to add station with existing seed.

JasonLG1979:
As mentioned in https://github.com/pithos/pithos/issues/385 Pandora does not allow multiple stations with the same seed. If you try to create a station with an existing seed it will fail and Pandora will just return the station containing the exsisting seed. currently we do not check for this and we blindly just add the station to the station model, which if it already exsists leads to a duplicate station. This fixes that by checking if the returned station already exsists and if it does it pops up a dialog telling the user that a new station could not be created and gives them the option to listen to the exsisting station that contains the seed they were trying to use to create a new station with.

@TingPing I'm not married to the wording in the dialog. I'm open to changes.

### DIFF
--- a/pithos/StationsDialog.py
+++ b/pithos/StationsDialog.py
@@ -12,6 +12,7 @@
 # You should have received a copy of the GNU General Public License along
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import html
 import logging
 
 from gi.repository import Gtk
@@ -38,6 +39,7 @@ class StationsDialog(Gtk.Dialog):
         self.worker_run = pithos.worker_run
         self.quickmix_changed = False
         self.searchDialog = None
+        self.result = None
 
         self.modelfilter = self.model.filter_new()
         self.modelfilter.set_visible_func(lambda m, i, d: m.get_value(i, 0) and not m.get_value(i, 0).isQuickMix)
@@ -153,15 +155,62 @@ class StationsDialog(Gtk.Dialog):
 
     @GtkTemplate.Callback
     def add_station_cb(self, dialog, response):
-        logging.info("in add_station_cb {} {}".format(dialog.result, response))
+        self.result = dialog.result
+        logging.info("in add_station_cb {} {}".format(self.result, response))
         if response == Gtk.ResponseType.OK:
-            self.worker_run("add_station_by_music_id", (dialog.result.musicId,),
+            self.worker_run("add_station_by_music_id", (self.result.musicId,),
                             self.station_added, "Creating station...")
         dialog.hide()
         dialog.destroy()
         self.searchDialog = None
 
+    def station_already_exists(self, station):
+        def on_response(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.pithos.station_changed(station)
+            dialog.destroy()
+
+        sub_title = 'Pandora does not permit multiple stations with the same seed.\n'
+
+        if self.result.resultType is 'song':
+            seed = 'Song Seed: {} by {}'.format(html.escape(self.result.title), html.escape(self.result.artist))
+        else:
+            seed = 'Artist Seed: {}'.format(html.escape(self.result.name))
+
+        self.result = None
+
+        if station is self.pithos.current_station:
+            button_type = Gtk.ButtonsType.OK
+            message = (
+                '{}"{}", the Station you are currently listening '
+                'to already contains the {}.'.format(sub_title, station.name, seed)
+            )
+
+        else:
+            button_type = Gtk.ButtonsType.YES_NO
+            message = (
+                '{}Your Station "{}" already contains the {}.\n'
+                'Would you like to listen to it now?'.format(sub_title, station.name, seed)
+            )
+
+        dialog = Gtk.MessageDialog(
+            parent=self,
+            flags=Gtk.DialogFlags.MODAL,
+            type=Gtk.MessageType.WARNING,
+            buttons=button_type,
+            text='A New Station could not be created',
+            secondary_text=message,
+        )
+
+        dialog.connect('response', on_response)
+        dialog.show()
+
     def station_added(self, station):
+        for existing_station in self.model:
+            if existing_station[0].id == station.id:
+                self.station_already_exists(existing_station[0])
+                return
+        self.result = None
         logging.debug("1 " + repr(station))
         it = self.model.insert_with_valuesv(0, (0, 1, 2), (station, station.name, 0))
         logging.debug("2 " + repr(it))


pithos/StationsDialog.py
```diff
@@ -153,15 +155,41 @@ def refresh_stations(self, widget):
...
+            seed = 'Song Seed: {} by {}'.format(html.escape(self.result.title), html.escape(self.result.artist))
+        else:
+            seed = 'Artist Seed: {}'.format(html.escape(self.result.name))
+        dialog.set_property(
+            'text',
+            'A New Station could not be created.\n'
+            'The Station "{}" already exists with the {}.\n'
+            'Would you like to listen to it now?'.format(station.name, seed)
+        )
+        response = dialog.run()
```
TingPing: Never use `Gtk.Dialog.run()`, it creates a new mainloop and causes all sorts of weird behavior.

pithos/StationsDialog.py
```diff
@@ -153,15 +155,41 @@ def refresh_stations(self, widget):
...
         if response == Gtk.ResponseType.OK:
-            self.worker_run("add_station_by_music_id", (dialog.result.musicId,),
+            self.worker_run("add_station_by_music_id", (self.result.musicId,),
                             self.station_added, "Creating station...")
         dialog.hide()
         dialog.destroy()
         self.searchDialog = None
 
+    def station_already_exists(self, station):
+        # May as well reuse the delete_confirm_dialog. All we need is some text and a Yes, No.
```
TingPing: I think this is very unclear behavior (delete dialog showing add warning?) just make a new dialog and let it be destroyed. The same should be done for a few of the other dialogs in the future IMO.

JasonLG1979: I can do that but I'm not sure who you're confusing? The user doesn't know the difference. delete_confirm_dialog is just a name. It's just an empty message dialog. When shown on screen it has no title. It's just a box with text and 2 buttons.

pithos/StationsDialog.py
JasonLG1979: OK

pithos/StationsDialog.py
JasonLG1979: So do you want me to make a UI file for basically an empty dialog or skip the UI file and write it all in python?

TingPing: Code is read by humans too :P

It can probably done in pure Python, it is little more than a default Gtk.MessageDialog.

JasonLG1979: OK not a problem. I'll get right on it.

JasonLG1979:
Something like that? It now also checks to see if the duplicate station is the current station.

JasonLG1979:
Do you think I need to include a better explanation as to why the station could not be added? Something like  "A New Station could not be created. Pandora does not permit multiple stations with the same seed."

pithos/StationsDialog.py
```diff
@@ -153,15 +155,63 @@ def refresh_stations(self, widget):
...
+    def station_already_exists(self, station):
+        def on_responce(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.pithos.station_changed(station)
+            dialog.destroy()
+
+        title = (
+            'A New Station could not be created'
+        )
+
```
TingPing: __[flake8]__

pithos/StationsDialog.py
```diff
@@ -153,15 +155,58 @@ def refresh_stations(self, widget):
...
+        sub_title = 'Pandora does not permit multiple stations with the same seed.\n'
+
+        if self.result.resultType is 'song':
+            seed = 'Song Seed: {} by {}'.format(html.escape(self.result.title), html.escape(self.result.artist))
+        else:
+            seed = 'Artist Seed: {}'.format(html.escape(self.result.name))
+        self.result = None
+        if station is self.pithos.current_station:
+            button_type = Gtk.ButtonsType.OK
+            message = '"{}", the Station you are currently listening to already contains the {}.'.format(station.name, seed)
```
TingPing: __[flake8]__
* [E501] line too long (124 > 120 characters)

JasonLG1979:
OK I think I'm done messing with this. What do you think about the wording?

JasonLG1979:
Somewhat off point but what do you think about adding an edit stations button to the stations popover menu? Seems like a natural place for one.

TingPing:
> Somewhat off point but what do you think about adding an edit stations button to the stations popover menu? 

At some point I was thinking the dialog could be entirely replaced by the popover. I never thought of any designs I liked and I'm not sure the result would be better but maybe just adding a few features to it would be nice.

JasonLG1979:
> At some point I was thinking the dialog could be entirely replaced by the popover.

I think a maybe adding delete and rename to the right click menu and a button that opens the stations dialog.

Here's a crude mock up of how they might look combined. As you can see it gets messy real quick.
screenshot from 2017-01-15 15-05-29

At this point I'm not a fan of totally combining them.

TingPing:
Right click entries are an easy improvement for sure. Extra buttons might need some thought.

JasonLG1979:
After/if this PR is merged and I add the ability to add stations in the song treeview I'll get on it. On a side note. In there somewhere I'd like to add the ability to add user_data to the worker. It would be really handy. In this PR for example I have an instance var self.result that's really unnecessary if I could send it as user_data to the worker I wouldn't need it.

JasonLG1979:
```python
class GObjectWorker:

    def send(self, command, args=(), callback=None, errorback=None, user_data=None):
        def run(data):
            command, args, callback, errorback, user_data = data
            try:
                result = command(*args)
                if callback:
                    if user_data:
                        result = result, user_data 
                    GLib.idle_add(callback, result)
            except Exception as e:
                e.traceback = traceback.format_exc()
                if errorback:
                    GLib.idle_add(errorback, e)
        if errorback is None:
            errorback = self._default_errorback
        data = command, args, callback, errorback, user_data
        thread = threading.Thread(target=run, args=(data,))
        thread.daemon = True
        thread.start()

    def _default_errorback(self, error):
        logging.error("Unhandled exception in worker thread:\n{}".format(error.traceback))
```

TingPing:
Isn't that the point of the `args` kwarg?

pithos/StationsDialog.py
```diff
@@ -153,15 +155,63 @@ def refresh_stations(self, widget):
...
+        dialog = Gtk.MessageDialog(
+            parent=self,
+            flags=Gtk.DialogFlags.MODAL,
+            type=Gtk.MessageType.WARNING,
+            buttons=button_type,
+            text=title,
+            secondary_text=message,
+        )
+
+        dialog.connect('response', on_responce)
```
TingPing: Typo

JasonLG1979: I see I'll fix it here in a bit

JasonLG1979:
> Oh I see you want it in the result not for the function, sure.

Yep, I just want it along for the ride. Like I said it'll be nice not to have a bunch of single use instance vars muddying things up.

JasonLG1979:
I think I got the typo?

JasonLG1979:
While you're in a merging mood...

TingPing:
> While you're in a merging mood...

Ah you just caught me going to sleep, man I'm exhausted after making that late night patch! =)

JasonLG1979:
My bad, lol... It's only just after 9 where I'm at.

PR #397, opened 2017-01-16
Include genre stations in search

JasonLG1979:
The title pretty much sums it up ;P

### DIFF
--- a/pithos/SearchDialog.py
+++ b/pithos/SearchDialog.py
@@ -65,6 +65,8 @@ class SearchDialog(Gtk.Dialog):
                     mk = '<b>{}</b> by {}'.format(html.escape(i.title), html.escape(i.artist))
                 elif i.resultType is 'artist':
                     mk = '<b>{}</b> (artist)'.format(html.escape(i.name))
+                elif i.resultType is 'genre':
+                    mk = '<b>{}</b> (genre)'.format(html.escape(i.stationName))
                 self.model.append((i, mk))
             self.treeview.show()
         self.worker_run('search', (self.query,), callback, "Searching...")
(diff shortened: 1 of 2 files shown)


JasonLG1979:
Crap forgot to switch branches... Hold on. I'll fix it.

TingPing:
Good stuff!, I feel like we could improve sorting but I'll leave that for a later change.

JasonLG1979:
> I feel like we could improve sorting but I'll leave that for a later change.

I had to add 'includeNearMatches' to broaden the search results otherwise for example if you entered 80s you wouldn't get any genre results. With includeNearMatches you get some 80's based stations. I'll look into maybe throwing away some of the really low score artist and song results maybe?

JasonLG1979:
For example with "Korn" as a search I ofc get the band Korn but I also get a bunch a other crap with a score lower than 75. I think for artist and song results we could thin out the results a bit by setting 75 as our threshold.

JasonLG1979:
Another prop better idea is to let the user decide what they're searching for Song, Artist or Genre and only show the results they want.

TingPing:
Yea some filter for type would be nice and yes we can probably throw away some really far off results.

JasonLG1979:
Something like:
```python
        l =  [SearchResult('artist', i) for i in results['artists']]
        l += [SearchResult('song',   i) for i in results['songs'] if i['score'] >= 80]
        l += [SearchResult('genre', i) for i in results['genreStations'] if i['score'] >= 80]
        l.sort(key=lambda i: i.score, reverse=True)
```

Will get rid of a lot of the noise.

JasonLG1979:
I can do a quick PR for that bit if you'd like?

JasonLG1979:
Do you want separate filter buttons or just one button that cycles though the filters , all, artist, song, genre?

TingPing:
Probably a drop down radio menu.

PR #407, opened 2017-01-24
Try to unlock the keyring if it's locked.

JasonLG1979:
At startup if the default keyring is locked try to get the user to unlock it. If the user does not unlock the keyring use the session collection. Passwords saved in the session collection do not survive past the login session.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -42,7 +42,7 @@ from .gobject_worker import GObjectWorker
 from .pandora import *
 from .pandora.data import *
 from .plugin import load_plugins
-from .util import parse_proxy, open_browser, get_account_password, popup_at_pointer
+from .util import parse_proxy, open_browser, get_account_password, popup_at_pointer, unlock_keyring
 
 try:
     import pacparser
@@ -161,18 +161,23 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.set_proxy(reconnect=False)
         self.set_audio_quality()
 
-        email = self.settings['email']
         try:
+            unlock_keyring()
+            email = self.settings['email']
             password = get_account_password(email)
         except GLib.Error as e:
             if e.code == 2:
-                self.fatal_error_dialog(e.message, _('You need to install a service such as gnome-keyring.'))
+                logging.error('You need to install a service such as gnome-keyring. Error: {}'.format(e))
+                self.fatal_error_dialog(
+                    e.message,
+                    _('You need to install a service such as gnome-keyring.'),
+                )
+
+        if not email or not password:
+            self.show()
+            self.show_preferences()
         else:
-            if not email or not password:
-                self.show()
-                self.show_preferences()
-            else:
-                self.pandora_connect()
+            self.pandora_connect()
 
     def init_core(self):
         #                                Song object            display text  icon  album art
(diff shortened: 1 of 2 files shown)


pithos/pithos.py
```diff
@@ -163,16 +163,17 @@ def __init__(self, app, test_mode):
 
         email = self.settings['email']
         try:
-            password = get_account_password(email)
+            test = get_account_password(email)
```
* [F841] local variable 'test' is assigned to but never used

pithos/util.py
```diff
@@ -27,8 +27,35 @@
 _ACCOUNT_SCHEMA = Secret.Schema.new('io.github.Pithos.Account', Secret.SchemaFlags.NONE,
                                     {"email": Secret.SchemaAttributeType.STRING})
 
+_current_collection = Secret.COLLECTION_DEFAULT
 
 # TODO: Async
+def unlock_keyring():
```
TingPing: __[flake8]__

JasonLG1979:
@TingPing I this isn't perfect. I'm unsure of the best way to catch the GLib.Error to show the fatal_error_dialog?

JasonLG1979:
Never mind @TingPing, I fired up Kubuntu 16.04 in a VM and found the failure point.

JasonLG1979:
This is kinda a major bug fix...

pithos/pithos.py
```diff
@@ -161,18 +161,20 @@ def __init__(self, app, test_mode):
         self.set_proxy(reconnect=False)
         self.set_audio_quality()
 
+        keyring_error = unlock_keyring()
```
TingPing: Instead of having it return an error just let the exception propagate and handle it here?

JasonLG1979:
It's easier to return the error to Pithos.py. That's where the error dialog is.

JasonLG1979:
Do you mean you mean a try except there instead of in the function? I though it would be cleaner to catch it in the function where I know exactly where it's going to break if it does.

TingPing:
Returning errors is not pythonic. While I hate exceptions it is the way it is intended to be used.

JasonLG1979:
Ok I'll amend it in a bit

JasonLG1979:
Something like that?

pithos/pithos.py
```diff
@@ -161,18 +161,23 @@ def __init__(self, app, test_mode):
...
             if e.code == 2:
-                self.fatal_error_dialog(e.message, _('You need to install a service such as gnome-keyring.'))
+                logging.error('You need to install a service such as gnome-keyring. Error: {}'.format(e))
+                self.fatal_error_dialog(
+                    e.message,
+                    _('You need to install a service such as gnome-keyring.'),
+                )
+
+        email = self.settings['email']
+        password = get_account_password(email)
```
TingPing: Still possible for this to raise an exception, so should still handle it too.

JasonLG1979: I just tested in a Kubuntu 16.04 VM without gnome-keyring installed. It caught the error and showed the fatal dialog. I once we get past the unlocking of the keyring I don't think anything else should raise an exception. We access the keyring when we try to unlock it if anything were to happen it will happen then. But If you want I can include that in the try except block?

JasonLG1979:
That work for you?

pithos/pithos.py
TingPing: > Once we get past the unlocking of the keyring I don't think anything else should raise any exceptions.

With any external resource you have to remember that your running a multi-process system, between two calls in your program the keyring can change state. This is uncommon of course but there is no reason to not handle edge cases.

PR #411, opened 2017-01-26
transient song notifications for GNOME Shell 3.20+

JasonLG1979:
GNOME Shell 3.20 or higher has bultin MPRIS functionality that makes persistent song notifications redundant.

### DIFF
--- a/pithos/plugins/notify.py
+++ b/pithos/plugins/notify.py
@@ -45,6 +45,18 @@ class NotifyPlugin(PithosPlugin):
         if 'action-icons' in caps:
             self.notification.set_hint('action-icons', GLib.Variant('b', True))
 
+        # GNOME Shell 3.20 or higher has bultin MPRIS functionality that makes
+        # persistent song notifications redundant.
+        has_built_in_mpris = False
+        version = server_info['version'].split('.')
+        if server_info['name'] == 'gnome-shell' and len(version) >= 2:
+            major_version, minor_version = (int(x) if x.isdigit() else 0 for x in version[0:2])
+            if major_version == 3 and minor_version >= 20:
+                has_built_in_mpris = True
+
+        if 'persistence' in caps and has_built_in_mpris:
+            self.notification.set_hint('transient', GLib.Variant('b', True))
+
         server_info = '\n'.join(('{}: {}'.format(k, v) for k, v in server_info.items()))
         logging.debug('\nNotification Server Information:\n{}'.format(server_info))
 


pithos/plugins/notify.py
```diff
@@ -45,6 +45,14 @@ def on_prepare(self):
         if 'action-icons' in caps:
             self.notification.set_hint('action-icons', GLib.Variant('b', True))
 
+        # GNOME Shell 3.20 or higher has bultin MPRIS functionality that makes
+        # persistent song notifications redundant.
+        min_version = Gtk.get_major_version() >= 3 and Gtk.get_minor_version() >= 20
```
TingPing: Gtk version doesn't necessarily mean Gnome-Shell version. Maybe there is some better way to access this, note that we'll need a way that works within the flatpak sandbox too.

JasonLG1979: > Gtk version doesn't necessarily mean Gnome-Shell version.

I didn't think it was possible to run a mismatched Gtk and GNOME Shell version?

> note that we'll need a way that works within the flatpak sandbox too

I was under the impression that only Gio.Notifications worked in the sandbox?

JasonLG1979: > Maybe there is some better way to access this

The server_info contains a version. But again that is the version of the server not necessarily the version of GNOME Shell.

TingPing: > I didn't think it was possible to run a mismatched Gtk and GNOME Shell version?

Sure it is, especially with the previously mentioned Flatpak it will become more common.

> The server_info contains a version. But again that is the version of the server not necessarily the version of GNOME Shell.

The MPRIS server is in gnome-shell, so I think that will do.

JasonLG1979: > The MPRIS server is in gnome-shell, so I think that will do

I mean the notification server version not MPRIS.

TingPing: Yea sorry, I mis-typed.

JasonLG1979: I see what you're getting at though. Gtk version would get the version of the bundled Gtk and the server version would get you the Gtk/shell version of the host

TingPing: Indeed.

pithos/plugins/notify.py
```diff
@@ -45,6 +45,23 @@ def on_prepare(self):
...
+        # persistent song notifications redundant.
+        has_built_in_mpris = False
+        if server_info['name'] == 'gnome-shell':
+            version = server_info['version'].split('.')
+            if version and len(version) >= 2:
+                try:
+                    major_version = int(version[0])
+                    minor_version = int(version[1])
+                except ValueError:
+                    major_version = minor_version = 0
```
TingPing: __[flake8]__
* [W291] trailing whitespace

JasonLG1979:
Something like that?

pithos/plugins/notify.py
```diff
@@ -45,6 +45,23 @@ def on_prepare(self):
...
+                if major_version >= 3 and minor_version >= 20:
```
TingPing: To future proof, 4.0 should probably be true.

JasonLG1979: 4.0 would represent a total change in API/ABI in GNOME versioning,  much like 2.x to 3.0. More than likely the whole app would need to be ported to 4.0.

JasonLG1979: I can make 4.0 equal true but I don't know if that would be more harm than good really. The way GNOME changes things from release to release even in minor versions who knows what 4.0 will actually mean?

TingPing: Perhaps, maybe make it major_version == 3 to be explicit then?

JasonLG1979: Will do

JasonLG1979:
Something like that?

pithos/plugins/notify.py
```diff
@@ -45,6 +45,18 @@ def on_prepare(self):
         if 'action-icons' in caps:
             self.notification.set_hint('action-icons', GLib.Variant('b', True))
 
+        # GNOME Shell 3.20 or higher has bultin MPRIS functionality that makes
+        # persistent song notifications redundant.
+        has_built_in_mpris = False
+        version = server_info['version'].split('.')
```
TingPing: is version guaranteed to be present?

JasonLG1979: If the server follows the spec. Otherwise I would expect they would return empty strings if all else fails.

JasonLG1979: org.freedesktop.Notifications.GetServerInformation
void org.freedesktop.Notifications.GetServerInformation (	name,	 
 	vendor,	 
 	version,	 
 	spec_version);	 
out STRING name;
out STRING vendor;
out STRING version;
out STRING spec_version;

JasonLG1979: It doesn't say anything about sending an error back if it fails?

JasonLG1979: I've never seen it fail. If it had we would have got bug reports about GioNotify I'd think since it turns the list of the values into the dict.

TingPing: Sounds fine.

PR #414, opened 2017-01-30
Stop hard-coding the logging level for journald

stevekuznetsov:
Today it is not possible to pass a more nuanced logging level to the
journald logger by starting Pithos using `--debug` or `--verbose`. This
change allows for the same logging level to be used for the journald
logger as for the stream logger.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

### DIFF
--- a/pithos/application.py
+++ b/pithos/application.py
@@ -91,6 +91,14 @@ class PithosApplication(Gtk.Application):
             type(command_line).do_print_literal(command_line, "Pithos {}\n".format(self.version))
             return 0
 
+        # Set the logging level to show debug messages
+        if options.contains('debug'):
+            log_level = logging.DEBUG
+        elif options.contains('verbose'):
+            log_level = logging.INFO
+        else:
+            log_level = logging.WARN
+
         handlers = []
         try:
             from systemd.journal import JournalHandler
@@ -99,20 +107,12 @@ class PithosApplication(Gtk.Application):
 
             # We can be more verbose with the journal and filter it later
             # and don't need fancy formatting as its part of the structure
-            journal.setLevel(logging.INFO)
+            journal.setLevel(log_level)
             journal.setFormatter(logging.Formatter())
 
             handlers.append(journal)
         except ImportError:
             pass
-
-        # Set the logging level to show debug messages
-        if options.contains('debug'):
-            log_level = logging.DEBUG
-        elif options.contains('verbose'):
-            log_level = logging.INFO
-        else:
-            log_level = logging.WARN
 
         stream = logging.StreamHandler()
         stream.setLevel(log_level)


stevekuznetsov:
@JasonLG1979 this should allow for an easier workflow with `journald`

JasonLG1979:
I'd like there to be a dropdown in the advanced settings in the prefs window that would allow you to change the logging level to the journal on the fly.

stevekuznetsov:
Setting the logging level at run-time would be absolutely amazing. Would you expect passing `--debug` to not affect output to `journald`, though, or would you expect `--debug` and `--verbose` to, if passed, change the menu item from the get-go?

JasonLG1979:
Idk how it all works @TingPing would be the person to talk to. I'm not sure what the norm in that situation would be but I would think that a command line arg would override the pref setting but that would lead to inconsistency between what the UI says and what's actually going on.

JasonLG1979:
Maybe gray-out and make unclickable the pref setting if a command line arg is provided?

JasonLG1979:
I don't like the idea of the arg actually changing the pref value though.

JasonLG1979:
> The entire point of the Journald logging is for desktop users, so they can't pass args.

I haven't tested his PR put it kinda does look like the command line args effect what get's logged by Journald?

JasonLG1979:
I guess my question is can you change the Journald log level at run time? I don't expect it to be retroactive.

TingPing:
If you run `pithos` from the command line, such as one would wanted verbose debug output, then they can pass the args and that is fine. The journal logging is filterable but not configurable so you want to be verbose enough common issues are logged but not verbose enough you fill their hard drive. No you can't retroactively change it.

JasonLG1979:
So you can't change the filter at run time?

JasonLG1979:
I guess my question is once set you can't change journal.setLevel?

TingPing:
You can change it after running but I don't get the point. The journal logging was added for this use case:

- User reports an issue
- We ask them, paste the logs from the journal when that issue happened

For that use-case changing it after that fact is a moot point because we already have this workflow:

- User reports an issue
- Ask them to run `pithos --debug` and paste it

JasonLG1979:
It would be useful for issues that only seem to pop up when your *not* running with pithos --debug

TingPing:
Sure but the problem with always logging debug is it is just extremely verbose and the journal is saved to disk wasting space.

JasonLG1979:
> Sure but the problem with always logging debug is it is just extremely verbose and the journal is saved to disk wasting space.

For sure. I wouldn't do it all the time. But if you know something might pop up it could be handy. We could put a disclaimer saying that debug logging is not recommended for normal use.

JasonLG1979:
I already have a rough working example.
screenshot from 2017-01-30 22-49-27

JasonLG1979:
https://github.com/JasonLG1979/pithos/tree/user-set-log-level

TingPing:
I'm fine with some setting to toggle it, not sure about adding it there in preferences though.

JasonLG1979:
Where would you put it then?

JasonLG1979:
Would you consider systemd logging a plugin?

TingPing:
Hmm, maybe it could be.. it is certainly optional.

JasonLG1979:
If you make it a plugin then the question is do you default it to enabled?

JasonLG1979:
Either way I like it as a plugin

TingPing:
> If you make it a plugin then the question is do you default it to enabled?

Seems reasonable.

JasonLG1979:
OK I'll see what I can do, but it would be nice if you merged a few of my PR's 1st ;P

PR #415, opened 2017-02-01
Make journald logging a plugin

JasonLG1979:
Users can now enable, disable and set the logging level of the journald logging in Pithos.

It's set to disabled by default like all plugins. I'm not sure if I want it defaulted to enabled?

Edit: It now defaults to enabled.

### DIFF
--- a/pithos/application.py
+++ b/pithos/application.py
@@ -49,6 +49,8 @@ class PithosApplication(Gtk.Application):
                              _('Use a mock service instead of connecting to the real Pandora server'), None)
         self.add_main_option('version', 0, GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
                              _('Show the version'), None)
+        self.add_main_option('last-logs', 0, GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
+                             _('Show the logs for Pithos since the last reboot'), None)
 
     def do_startup(self):
         Gtk.Application.do_startup(self)
@@ -85,26 +87,55 @@ class PithosApplication(Gtk.Application):
         # http://stackoverflow.com/questions/1943747/python-logging-before-you-run-logging-basicconfig
         logging.root.handlers = []
 
+        # Show the Pithos log since last reboot and exit
+        if options.contains('last-logs'):
+            try:
+                from systemd import journal
+                from os.path import basename
+            except ImportError:
+                self._print(command_line, _('Systemd Python module not found'))
+                return 1
+
+            # We want the version also since the logging plugin misses
+            # logging messages before it's enabled.
+            self._print(command_line, 'Pithos {}'.format(self.version))
+
+            reader = journal.Reader()
+            reader.this_boot()
+            reader.add_match(SYSLOG_IDENTIFIER='io.github.Pithos')
+
+            _PRIORITY_TO_LEVEL = {
+                journal.LOG_DEBUG: 'DEBUG',
+                journal.LOG_INFO: 'INFO',
+                journal.LOG_WARNING: 'WARNING',
+                journal.LOG_ERR: 'ERROR',
+                journal.LOG_CRIT: 'CRTICIAL',
+                journal.LOG_ALERT: 'ALERT',
+            }
+
+            got_logs = False            
+
+            for entry in reader:
+                got_logs = True
+                level = _PRIORITY_TO_LEVEL[entry['PRIORITY']]
+                line = entry['CODE_LINE']
+                function = entry['CODE_FUNC']
+                module = basename(entry['CODE_FILE'])[:-3]
+                message = entry['MESSAGE']
+                log_line = '{} - {}:{}:{} - {}'.format(level, module, function, line, message)
+                self._print(command_line, log_line)
+
+            if not got_logs:
+                self._print(command_line, _('No logs for Pithos present for this boot.'))
+
+            return 0
+
         # Show the version on local instance and exit
         if options.contains('version'):
-            # Broken bindings...
-            type(command_line).do_print_literal(command_line, "Pithos {}\n".format(self.version))
+            self._print(command_line, 'Pithos {}'.format(self.version))
             return 0
 
         handlers = []
-        try:
-            from systemd.journal import JournalHandler
-
-            journal = JournalHandler(SYSLOG_IDENTIFIER=self.props.application_id)
-
-            # We can be more verbose with the journal and filter it later
-            # and don't need fancy formatting as its part of the structure
-            journal.setLevel(logging.INFO)
-            journal.setFormatter(logging.Formatter())
-
-            handlers.append(journal)
-        except ImportError:
-            pass
 
         # Set the logging level to show debug messages
         if options.contains('debug'):
@@ -126,6 +157,11 @@ class PithosApplication(Gtk.Application):
         self.do_activate()
 
         return 0
+
+    @staticmethod
+    def _print(command_line, string):
+        # Workaround broken pygobject bindings
+        type(command_line).do_print_literal(command_line, string + '\n')
 
     def do_activate(self):
         if not self.window:
--- a/pithos/plugins/journald_logging.py
+++ b/pithos/plugins/journald_logging.py
@@ -1,0 +1,146 @@
+#
+# Copyright (C) 2016 Jason Gray <jasonlevigray3@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3, as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranties of
+# MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+# END LICENSE
+
+import logging
+
+from gi.repository import GObject, Gtk
+
+from pithos.plugin import PithosPlugin
+
+LOG_LEVELS = {
+    'debug': logging.DEBUG,
+    'verbose': logging.INFO,
+    'warning': logging.WARN,
+}
+
+
+class JournalLoggingPlugin(PithosPlugin):
+    preference = 'journald-logging'
+    description = _('Store logs with the journald service')
+
+    _logging_changed_handler = None
+
+    def on_prepare(self):
+        try:
+            from systemd.journal import JournalHandler
+            self._journal = JournalHandler(SYSLOG_IDENTIFIER='io.github.Pithos')
+            self._journal.setFormatter(logging.Formatter())
+            self._logger = logging.getLogger()
+            self.preferences_dialog = LoggingPluginPrefsDialog(self.window, self.settings)
+        except ImportError:
+            return _('Systemd Python module not found')
+
+    def on_enable(self):
+        self._on_logging_changed(None, self.settings['data'] or 'verbose')
+        self._logger.addHandler(self._journal)
+        self._logging_changed_handler = self.preferences_dialog.connect('logging-changed', self._on_logging_changed)
+
+    def _on_logging_changed(self, prefs_dialog, level):
+        self.settings['data'] = level
+        self._journal.setLevel(LOG_LEVELS[level])
+        logging.info('setting journald logging level to: {}'.format(level))
+
+    def on_disable(self):
+        if self._logging_changed_handler:
+            self.preferences_dialog.disconnect(self._logging_changed_handler)
+        self._logger.removeHandler(self._journal)
+
+
+class LoggingPluginPrefsDialog(Gtk.Dialog):
+    __gtype_name__ = 'LoggingPluginPrefsDialog'
+    __gsignals__ = {
+        'logging-changed': (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_STRING,)),
+    }
+
+    def __init__(self, parent, settings, *args, **kwargs):
+        super().__init__(
+            _('Logging Level'),
+            parent,
+            0,
+            ('_Cancel', Gtk.ResponseType.CANCEL, '_Apply', Gtk.ResponseType.APPLY),
+            *args,
+            use_header_bar=1,
+            **kwargs,
+        )
+        self.pithos_window = parent
+        self.settings = settings
+        self.set_resizable(False)
+        self.connect('response', self._on_response)
+
+        sub_title = Gtk.Label.new(_('Set the journald logging level for Pithos'))
+        sub_title.set_halign(Gtk.Align.CENTER)
+        self.log_level_combo = Gtk.ComboBoxText.new()
+
+        logging_levels = [
+            ('debug', 'High - debug'),
+            ('verbose', 'Default - verbose'),
+            ('warning', 'Low - warning'),
+        ]
+
+        for level in logging_levels:
+            self.log_level_combo.append(level[0], level[1])
+
+        self._reset_combo()
+        content_area = self.get_content_area()
+
+        content_area.add(sub_title)
+        content_area.add(self.log_level_combo)
+        content_area.show_all()
+
+    def _reset_combo(self):
+        self.log_level_combo.set_active_id(self.settings['data'] or 'verbose')
+
+    def _on_response(self, dialog, response):
+        if response != Gtk.ResponseType.APPLY:
+            self.hide()
+            self._reset_combo()
+            return
+
+        setting = self.settings['data']
+        active_id = self.log_level_combo.get_active_id()
+
+        if setting == active_id:
+            self.hide()
+            return
+
+        if active_id != 'debug':
+            self.hide()
+            self.emit('logging-changed', active_id)
+            return
+
+        def on_dialog_response(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.hide()
+                self.emit('logging-changed', active_id)
+            dialog.destroy()
+
+        message = (_(
+            'The debug logging level is not '
+            'recommended unless you are actually debugging an issue, '
+            'as it generates very large logs.\n\nAre you sure you want to set logging to debug?',
+        ))
+
+        dialog = Gtk.MessageDialog(
+            parent=self.pithos_window.prefs_dlg,
+            flags=Gtk.DialogFlags.MODAL,
+            type=Gtk.MessageType.WARNING,
+            buttons=Gtk.ButtonsType.YES_NO,
+            text=_('Debug Logging Level'),
+            secondary_text=message,
+        )
+
+        dialog.connect('response', on_dialog_response)
+        dialog.show()
--- a/data/io.github.Pithos.gschema.xml
+++ b/data/io.github.Pithos.gschema.xml
@@ -66,6 +66,7 @@
     <child name="mediakeys" schema="io.github.Pithos.plugin-enabled"/>
     <child name="screensaver-pause" schema="io.github.Pithos.plugin-enabled"/>
     <child name="mpris" schema="io.github.Pithos.plugin-enabled"/>
+    <child name="journald-logging" schema="io.github.Pithos.plugin-enabled"/>
     <child name="notify" schema="io.github.Pithos.plugin"/>
     <child name="lastfm" schema="io.github.Pithos.plugin"/>
     <child name="notification-icon" schema="io.github.Pithos.plugin"/>
--- a/pithos/PreferencesPithosDialog.py
+++ b/pithos/PreferencesPithosDialog.py
@@ -50,12 +50,23 @@ class PithosPluginRow(Gtk.ListBoxRow):
         self.switch.connect('notify::active', self.on_activated)
         self.switch.set_valign(Gtk.Align.CENTER)
         box.pack_end(self.switch, False, False, 2)
+        self.connect('grab-focus', self.on_grab_focus)
 
         if plugin.prepared and plugin.error:
             self.set_sensitive(False)
             self.set_tooltip_text(plugin.error)
 
         self.add(box)
+
+    def set_prefs_btn(self):
+        if self.plugin.enabled:
+            sensitive = self.plugin.preferences_dialog is not None
+        else:
+            sensitive = False
+        self.get_toplevel().preference_btn.set_sensitive(sensitive)
+
+    def on_grab_focus(self, *ignore):
+        self.set_prefs_btn()
 
     def on_activated(self, obj, params):
         if not self.is_selected():
@@ -71,7 +82,7 @@ class PithosPluginRow(Gtk.ListBoxRow):
             self.set_sensitive(False)
             self.set_tooltip_text(self.plugin.error)
         elif self.plugin.prepared:
-            self.get_toplevel().preference_btn.set_sensitive(self.plugin.preferences_dialog is not None)
+            self.set_prefs_btn()
 
 
 @GtkTemplate(ui='/io/github/Pithos/ui/PreferencesPithosDialog.ui')
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -11,4 +11,5 @@ data/io.github.Pithos.desktop.in
 pithos/application.py
 pithos/pithos.py
 pithos/StationsPopover.py
+pithos/plugins/journald_logging.py
 


TingPing:
> It's set to disabled by default like all plugins. I'm not sure if I want it defaulted to enabled?

Yea we need to fix that.

pithos/plugins/journald_logging.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def on_prepare(self):
+        try:
+            from systemd.journal import JournalHandler
+            self._journal = JournalHandler(SYSLOG_IDENTIFIER='io.github.Pithos')
+            self._journal.setFormatter(logging.Formatter())
+            self._logger = logging.getLogger()
+            self.preferences_dialog = LoggingPluginPrefsDialog(self.window, self.settings)
+        except ImportError:
+            return 'Journald Logging Service not found'
```
TingPing: `'Systemd Python module not found'`

stevekuznetsov:
I'm not sure I understand why it is better to import the `journald` driver and log to it instead of writing a `systemd` unit file and letting this feature just change the logging level at run-time, where `systemd` would just consume all of the output from the program regardless.

TingPing:
The automatic stdout handling of systemd loses information such as log levels, module and line that created the log, etc. Using the module we can set any arbitrary structured data we wish.

JasonLG1979:
> Yea we need to fix that.

How would we do that? Have a seperate generic plugin gsetting gor plugs that default to enabled?

pithos/plugins/journald_logging.py
JasonLG1979: No problem. I'll have to do a little git magic change that commit.

JasonLG1979:
Something like this maybe?
```python
def load_plugins(window):
    plugins = window.plugins

    enabled_by_default = [
        'journald_logging'
    ]

    plugins_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "plugins")
    discovered_plugins = (fname[:-3] for fname in glob.glob1(plugins_dir, "*.py") if not fname.startswith("_"))

    for name in discovered_plugins:
        if name not in plugins:
            plugin = plugins[name] = load_plugin(name, window)
        else:
            plugin = plugins[name]
        if name in enabled_by_default:
            schema_id = 'io.github.Pithos.plugin-default-enabled'
        else:
            schema_id = 'io.github.Pithos.plugin'        
        plugin.settings = Gio.Settings.new_with_path(schema_id, '/io/github/Pithos/{}/'.format(name))

        if plugin.settings['enabled']:
            plugin.enable()
        else:
            plugin.disable()
```

stevekuznetsov:
> The automatic stdout handling of systemd loses information such as log levels, module and line that created the log, etc. Using the module we can set any arbitrary structured data we wish.

@TingPing I see, this makes sense.

JasonLG1979:
> The automatic stdout handling of systemd loses information such as log levels, module and line that created the log, etc. Using the module we can set any arbitrary structured data we wish.

We currently don't do anything as far as formating:
https://github.com/pithos/pithos/blob/master/pithos/application.py#L103

Would you like the journald logging format to match the stream logger?

JasonLG1979:
@TingPing There now the logging plugin is enabled by default. And the format matches the stream logger.

TingPing:
>  And the format matches the stream logger.

Undo that part, there was even a comment saying it shouldn't before. Journald uses structures not formatting to include information. So in the log you have:

```
{
    'module': 'pithos.py',
    'level': 'WARNING',
     'message: 'Something broke!'
}
```

So you can see how it is redundant to include that in the message.

JasonLG1979:
I get none of that when I view the log in Logs. or dump it to file. All I get is the time program name and message?

``` 6:51:52 PM pithos: Downloaded album art for 3```

screenshot from 2017-02-01 23-14-48

**Comment by TingPing** [comment]
`gnome-logs` is honestly trash. As a quick example though: `journalctl -t io.github.Pithos -o json-pretty`, you can ofc filter by specific fields though.

**Comment by JasonLG1979** [comment]
So have the user reconstruct log messages from a json. That's useful.

**Comment by TingPing** [comment]
Well generally they just pastebin a blob of text, if actually necessary we can make a script to format it for them or something.

**Comment by JasonLG1979** [comment]
It's so much easier if the message contains all the info. It's so much easier to have a user open a default program and copy and paste from the file dump.

**Comment by JasonLG1979** [comment]
I don't know about you but I really don't want to convert 200 json object into a readable log everything someone post an issue

**Comment by stevekuznetsov** [comment]
@JasonLG1979 so you can get a flattened view out of the system journal like so:
```
# journalctl -t io.github.Pithos --no-pager | head -n 10
-- Logs begin at Wed 2016-12-14 13:01:19 EST, end at Thu 2017-02-02 07:30:04 EST. --
Dec 14 16:25:46 io.github.Pithos[9742]: Selecting station 2833412658794023199; total = 49
Dec 14 16:25:46 io.github.Pithos[9742]: pandora: Get Playlist
Dec 14 16:25:52 io.github.Pithos[9742]: Selecting station 3172841149648029983; total = 49
Dec 14 16:25:52 io.github.Pithos[9742]: pandora: Get Playlist
Dec 14 16:25:53 io.github.Pithos[9742]: Network error: <urlopen error [Errno 104] Connection reset by peer>
Dec 14 16:25:58 io.github.Pithos[9742]: Manual retry
Dec 14 16:25:58 io.github.Pithos[9742]: pandora: Get Playlist
Dec 14 16:25:58 io.github.Pithos[9742]: Network error: <urlopen error EOF occurred in violation of protocol (_ssl.c:600)>
Dec 14 16:25:59 io.github.Pithos[9742]: Network error: <urlopen error [Errno 104] Connection reset by peer>
```

I snipped my hostname out of there but  it's usually in the output. the JSON output you can ask for with `-o json-pretty` among other options, which gives you things like:
```json
{
	"__CURSOR" : "s=0da1e44e8a484613a571d65846ddee3b;i=6ade1;b=e8679abb4c814163b7f2684b092fb85e;m=8072eee377;t=5475738857b53;x=ec3efdfb384479c",
	"__REALTIME_TIMESTAMP" : "1485814819552083",
	"__MONOTONIC_TIMESTAMP" : "551684072311",
	"_BOOT_ID" : "e8679abb4c814163b7f2684b092fb85e",
	"_UID" : "1000",
	"_GID" : "1001",
	"_CAP_EFFECTIVE" : "0",
	"_AUDIT_SESSION" : "2",
	"_AUDIT_LOGINUID" : "1000",
	"_SYSTEMD_CGROUP" : "/user.slice/user-1000.slice/session-2.scope",
	"_SYSTEMD_SESSION" : "2",
	"_SYSTEMD_OWNER_UID" : "1000",
	"_SYSTEMD_UNIT" : "session-2.scope",
	"_SYSTEMD_SLICE" : "user-1000.slice",
	"_SELINUX_CONTEXT" : "unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023",
	"_MACHINE_ID" : "232e8b1c7bc34b808c58345554828704",
	"_HOSTNAME" : "dhcp137-202.rdu.redhat.com",
	"CODE_FILE" : "/usr/local/lib/python3.4/site-packages/pithos/pithos.py",
	"SYSLOG_IDENTIFIER" : "io.github.Pithos",
	"LOGGER" : "root",
	"THREAD_NAME" : "MainThread",
	"_TRANSPORT" : "journal",
	"_PID" : "3980",
	"_COMM" : "pithos",
	"_EXE" : "/usr/bin/python3.5",
	"_CMDLINE" : "/usr/bin/python3 /usr/local/bin/pithos",
	"PRIORITY" : "3",
	"CODE_LINE" : "793",
	"CODE_FUNC" : "on_gst_error",
	"MESSAGE" : "Gstreamer error: gst-stream-error-quark: Stream doesn't contain enough data. (4), gsttypefindelement.c(985): gst_type_find_element_chain_do_typefinding (): /GstPlayBin:player/GstURIDecodeBin:uridecodebin330/GstTypeFindElement:typefindelement330:\nCan't typefind stream, 4",
	"_SOURCE_REALTIME_TIMESTAMP" : "1485814819548161"
}
```

"Users" are not asked to do anything with the raw data, but it is there to enable more complex use-cases. For instance, 

```
# journalctl -t io.github.Pithos --since "2 hours ago" --until "10 minutes ago"
```

You can't really do that with a flat string log. The binary nature of the logs is a huge boon. You could, for instance search through errors only from Pithos with:

```
# journalctl -t io.github.Pithos PRIORITY=3
```

stevekuznetsov:
Or even you can search for only GStreamer errors:
```
# journalctl -t io.github.Pithos CODE_FUNC=on_gst_error
```

TingPing:
Indeed, the entire point of journald is to have rich structured data, throwing everything in a single string is a thing of the past with syslog. Yes it has the downside of requiring a tool to manipulate but it can be so much more useful than that.

Now we can argue the best way for us to get these from users to pastebin (like i said we can make a simple command for this like `pithos --last-logs`) but there is no point in us mis-using the logging by throwing everything in the message.

JasonLG1979:
My point is that that no matter what even with all that command line fu you never end up with anything that resembles the output of the stream logger.

JasonLG1979:
It's of limited use if you have to have a user become an expert in systemd logging to report issues. It's so much easer to tell the user to open Logs and search for Pithos and dump it to file.

stevekuznetsov:
@JasonLG1979 it's not that it looks different, but I can ask for all the errors from a specific function in a specific file between two dates ... which, with a flat log, is frankly impossible without some really hard work. I didn't realize we gained that much metadata from the python interface to the system log and now that i know this, it makes a lot of sense. Using the system journal doesn't preclude have some textual logs somewhere if you want them, and for what it's worth in all the past bug reports all you've seen is nice flat messages, even though I may have been doing some command-line fu in the background to give you a more readable set of messages in the first place (like filtering by time).

JasonLG1979:
Ok what's the commands that will give me a flat log that resembles the stream logger?

TingPing:
Without external tools `journalctl` probably can't output the exact same format, which I don't see as a problem.

stevekuznetsov:
The output format from the system journal is changed with `--output` on `journalctl`. From the manual:
```
       -o, --output=
           Controls the formatting of the journal entries that are shown. Takes one of the following options:

           short
               is the default and generates an output that is mostly identical to the formatting of classic syslog files, showing one line per journal entry.

           short-iso
               is very similar, but shows ISO 8601 wallclock timestamps.

           short-precise
               is very similar, but shows timestamps with full microsecond precision.

           short-monotonic
               is very similar, but shows monotonic timestamps instead of wallclock timestamps.

           short-unix
               is very similar, but shows seconds passed since January 1st 1970 UTC instead of wallclock timestamps ("UNIX time"). The time is shown with microsecond accuracy.

           verbose
               shows the full-structured entry items with all fields.

           export
               serializes the journal into a binary (but mostly text-based) stream suitable for backups and network transfer (see Journal Export Format[1] for more information).

           json
               formats entries as JSON data structures, one per line (see Journal JSON Format[2] for more information).

           json-pretty
               formats entries as JSON data structures, but formats them in multiple lines in order to make them more readable by humans.

           json-sse
               formats entries as JSON data structures, but wraps them in a format suitable for Server-Sent Events[3].

           cat
               generates a very terse output, only showing the actual message of each journal entry with no metadata, not even a timestamp.
```

So basically, by default you get flat output with a simple `journalctl -t io.github.Pithos` you get the flat single-line output with timestamps, etc -- is that acceptable for your needs? If you wanted _only_ the messages you might need to use `jq` or something, but if you are OK with timestamps sitting in front of them then the default output will get you pretty close.

JasonLG1979:
You guys are conflating usability with functionality. True you can do a lot of fancy stuff with  journald but is the average user 1. Going to know how to manipulate the log like a ninja? 2. Going to really care that much to learn just to debug an issue in Pithos. 3. Do you really want to have to teach someone how to be log ninja EVERY time a noob wants to debug an issue?

stevekuznetsov:
If a user has an issue on a RHEL/CentOS/Fedora machine, they are either not accustomed with the platform or OK using the system journal. Getting paste-bin type logs is as simple as:

```
# journalctl -t io.github.Pithos --no-pager
```

JasonLG1979:
Again with nothing but the message...

stevekuznetsov:
Sorry, I'm not sure I understand -- why would it be bad for the user to also give us a timestamp? In any case, I was wrong before about just getting the message:

```
# journalctl -t io.github.Pithos -o cat --no-pager
```

JasonLG1979:
That means I have to search the code base to see where the message came from.

TingPing:
If we are talking about a 'average user' then the missing information like function it was from isn't useful in the first place. If we are talking about developers 1. The previous logging is untouched, 2. The journal gives you more information not less.

JasonLG1979:
I am talking the ease of talking an average user (I hate that term it's so vague) though how to get me useful log content. I do not have access the thier raw log so all the wiz bang features do no one any good.

JasonLG1979:
It's so much easier to tell the average user to open this app search for this, press this button and copy and paste this file than here enter this incompressible gibberish into the terminal and then copy and paste that. Then I get to analyze that output. It really a case that all the extra functionality actually gets in the way.

stevekuznetsov:
Wait -- the system journal log is just one of the logging handlers in the chain, right? The other logging is untouched, we just get this as well. I agree, if you are adding a "see logs" button it maybe should show just the messages / flat log .... and allow the user to copy the log file and/or the journald binary export to e-mail to a developer.

TingPing:
```python
from systemd import journal
from os.path import basename
import logging

reader = journal.Reader()
reader.this_boot()
reader.add_match(SYSLOG_IDENTIFIER='io.github.Pithos')

_PRIORITY_TO_LEVEL = {
    journal.LOG_DEBUG: 'DEBUG',
    journal.LOG_INFO: 'INFO',
    journal.LOG_WARNING: 'WARNING',
    journal.LOG_ERR: 'ERROR',
    journal.LOG_CRIT: 'CRTICIAL',
    journal. LOG_ALERT: 'ALERT',
}

for entry in reader:
    level = _PRIORITY_TO_LEVEL[entry['PRIORITY']]
    line = entry['CODE_LINE']
    function = entry['CODE_FUNC']
    module = basename(entry['CODE_FILE'])[:-3]
    message = entry['MESSAGE']
    print('{} - {}:{}:{} - {}'.format(level, module, function, line, message))
```

JasonLG1979:
My proposal as a compromise is adding a button to the plugin prefs window that dumps the Pithos logs to a readable file. If that's where you're going than cool I'm fine with no redundant stuff in the message.

TingPing:
I am totally fine with giving users a way to get that info, be it a command line flag or sure entries in the preferences window.

As @stevekuznetsov said the previously logging is still totally in place, this just gives us a better format to look back at previous information.

JasonLG1979:
It would be great to be able to say open the logging prefs, press this button and copy and paste from this file.

TingPing:
The one downside with it being in preferences is a fatal error can happen before they get that far. If it had a command line switch it could open its own window that doesn't interact with the rest of the application.

JasonLG1979:
Command line is fine. We should probably limit it though there maybe a lot of output. I've got logs for Pithos going back months

JasonLG1979:
As far as it's own window we really don't need that. They're already in the terminal may as well just print to the terminal and not even start Pithos.

TingPing:
Honestly this conversation has really made me want to make a not awful log viewer. :P

JasonLG1979:
> Honestly this conversation has really made me want to make a not awful log viewer. :P

Well you go for that. After you merge the like 10 latest PR's I have pending... Zing ;P

TingPing:
> Well you go for that. After you merge the like 10 latest PR's I have pending... Zing ;P

I am a terrible maintainer, I don't know why I keep landing in that position =(

JasonLG1979:
> I am a terrible maintainer

I like to think of you as being cautious. If you merged every hair brained idea I had Pithos would be sh!t.

TingPing:
I have dealt with a lot of shitty releases and code because I was over-zelous, so I have really slowed down in recent time. Part of my problem with Pithos PRs is that the codebase is still an untested mess, I cannot at a glance tell if any change you make to the player in `pithos.py` is actually sane.. =\

JasonLG1979:
> the codebase is still an untested mess

True story. I can change one thing and it will break in spectacularly weird and unexpected ways.

TingPing:
So that is my un-subtle hint towards you, if you want fast PRs modularize the player and make a test suite =)

JasonLG1979:
> from systemd import journal
> from os.path import basename
> 
> reader = journal.Reader()
> reader.this_boot()
> reader.add_match(SYSLOG_IDENTIFIER='io.github.Pithos')
> 
> _PRIORITY_TO_LEVEL = {
>     journal.LOG_DEBUG: 'DEBUG',
>     journal.LOG_INFO: 'INFO',
>     journal.LOG_WARNING: 'WARNING',
>     journal.LOG_ERR: 'ERROR',
>     journal.LOG_CRIT: 'CRTICIAL',
>     journal.LOG_ALERT: 'ALERT',
> }
> 
> for entry in reader:
>     level = _PRIORITY_TO_LEVEL[entry['PRIORITY']]
>     line = entry['CODE_LINE']
>     function = entry['CODE_FUNC']
>     module = basename(entry['CODE_FILE'])[:-3]
>     message = entry['MESSAGE']
>     print('{} - {}:{}:{} - {}'.format(level, module, function, line, message))

Doesn't output anything.

JasonLG1979:
Durp, never mind. I just realized it would help if I actually ran Pithos since I booted up today...

JasonLG1979:
@There you go. No redundant stuff in the log message and a command line arg to output the logs since the last reboot. Fair?

pithos/application.py
```diff
@@ -85,6 +87,38 @@ def do_command_line(self, command_line):
         # http://stackoverflow.com/questions/1943747/python-logging-before-you-run-logging-basicconfig
         logging.root.handlers = []
 
+        # Show the Pithos log since last reboot and exit
+        if options.contains('last-logs'):
+            try:
+                from systemd import journal
+                from os.path import basename
+            except ImportError:
+                type(command_line).do_print_literal(command_line, 'Systemd Python module not found')
```
TingPing: Missing a `\n` on all of these. Should move it to a helper function like so: https://github.com/lutris/lutris/blob/master/lutris/gui/application.py#L132-L135

JasonLG1979: OK. Also I noticed something. since it's a plugin it doesn't start logging until the plugin loads. I'm working on a hand off. Basically it will log like it does currently and when the logging plugin is loaded initial logging is disabled, and then picked back up by the logging plugin(if it's enabled)

JasonLG1979: On a closer look it really doesn't miss much anyway at the INFO level...Not really worth the trouble for a hand off.

JasonLG1979:
> So that is my un-subtle hint towards you, if you want fast PRs modularize the player and make a test suite =)

I'll make you a deal. If you review(and hopefully merge) my PR's 390 though this one so we can do a new release the next thing I do will be splitting out the player logic?

JasonLG1979:
@TingPing bump on this and ALL of my other PRs... Come on man either close or test and merge some stuff or I'm going to just close them all. I know you probability don't care and that's not a threat in the sense that I realize that closing my PRs doesn't hurt you in any way, it's just I'm tired of trying to make something better and nothing gets done. I'm starting to wonder why I even put in PRs? I might as well just keep my own little private branch and not even bother...

JasonLG1979:
This PR and the MPRIS Playlist support and I will be one happy camper, Thanks!!! Like I said I'd love to do a new release. I think the amount of fixes/changes warrants it, wouldn't you say?

TingPing:
Oops, I had no clue you hacked this to being enabled by default, well anyway you'll want to rebase it over 4af5426ca47d0000dcae5e0cd2204a2490297a4e

JasonLG1979:
My solution is better as it doesn't require you to hard code which plugins are enabled by default in the settings xml.

TingPing:
It just moves the hardcoding to the plugin. I think it is logically clearer that they exist in the settings file and while its a bummer to hardcode them it isn't like we actually have third party plugins yet.

JasonLG1979:
3rd party support wouldn't be hard and your way would break it. My way it's more future proof.

JasonLG1979:
The whole point of generic plugin settings is to avoid each plugin having it's own settings. Your way defeats that purpose.

TingPing:
Hmm, I'll fix it for third party ones but I do still think the defaults belongs in the schema file.

JasonLG1979:
But then you still have to have a way for 3rd party plugins to decide if they default on or not. It becomes may complicated. My way is so much simpler. The plugin puts default_enabled = True in there and it's default enabled otherwise it's not.

TingPing:
Why should a third party plugin have the ability to control it being loaded, that should be the users decision.

JasonLG1979:
It's a 3rd party plugin the user is the one who installed it. It's up to them to figure out how or even if it works not ours.

JasonLG1979:
3rd party plugins are sort of the wild west. Once we enable them what they do is not really our problem.

TingPing:
The end result will simply be all third party plugins will default on because they believe they deserve it. Also plugins are currently system wide so in theory users can share them and suddenly get a new plugin.

JasonLG1979:
You wouldn't make 3rd party plugins system wide. In my mind they would live somewhere in the users home. Pithos would look in a certain spot in the current user's home and load them from there. Why would you make 3rd party plugins system wide? That's just bonkers,lol!!!

TingPing:
I'm saying that is the current situation not that it is the ideal situation. Ideally I believe we should support both.

JasonLG1979:
> The end result will simply be all third party plugins will default on because they believe they deserve it.

Well that's up to the users of said 3rd party plugins to file bug reports with the creators of those plugins. Again not our problem.

> Ideally I believe we should support both.

Not me. In my mind the only plugins that should be system wide are the ones we include. 3rd party plugins should installed on a by user basis. If the 3rd party plugin was useful enough to be system wide they should put in a PR to have it included in the default plugins.

JasonLG1979:
We can't have unvetted 3rd party plugins from one user messing things up for other users.

TingPing:
So what do you think of this idea:

System path is is whitelisted for loading by default for our plugins

User path is simply always defaults to loading as the user did install it

JasonLG1979:
> So what do you think of this idea:
> 
> System path is is whitelisted for loading by default for our plugins
> 
> User path is simply always defaults to loading as the user did install it

No I still like my original solution. It's clean, it doesn't effect existing plugins. It's simple. It requires less code change. It doesn't hard code settings. And it makes no assumptions as to if the plugin should be loaded by default. IMHO Always loading 3rd party plugins by default is a horrible idea.

JasonLG1979:
@TingPing there you go all rebased on top of your changes. Should be ready to go.

pithos/application.py
```diff
@@ -85,26 +87,48 @@ def do_command_line(self, command_line):
...
+            _PRIORITY_TO_LEVEL = {
+                journal.LOG_DEBUG: 'DEBUG',
+                journal.LOG_INFO: 'INFO',
+                journal.LOG_WARNING: 'WARNING',
+                journal.LOG_ERR: 'ERROR',
+                journal.LOG_CRIT: 'CRTICIAL',
+                journal.LOG_ALERT: 'ALERT',
+            }
+
+            for entry in reader:
```
TingPing: Maybe we should print a message when no logs are found (like the plugin being disabled).

pithos/plugins/journald_logging.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+        self.settings = settings
+        self.set_resizable(False)
+        self.connect('response', self._on_response)
+
+        sub_title = Gtk.Label.new('Set the journald logging level for Pithos')
+        sub_title.set_halign(Gtk.Align.CENTER)
+
+        logging_levels = [
```
TingPing: Translate the label portion.

pithos/plugins/journald_logging.py
```diff
@@ -0,0 +1,150 @@
...
+            ('debug', 'High - debug'),
+            ('verbose', 'Default - verbose'),
+            ('warning', 'Low - warning'),
+        ]
+
+        fmt_store = Gtk.ListStore(GObject.TYPE_STRING, GObject.TYPE_STRING)
+        for logging_level in logging_levels:
+            fmt_store.append(logging_level)
+        self.log_level_combo = Gtk.ComboBox.new()
```
TingPing: Use `Gtk.ComboBoxText` perhaps.

pithos/application.py
JasonLG1979: OK makes sense. Not everyone would clue into the fact that it only outputs logs for that boot.(myself included,lol!!!)

pithos/plugins/journald_logging.py
JasonLG1979: OK.

Where and how do I add that to translations again?

TingPing: `_('string')` and add the file to `po/POTFILES.in`.

JasonLG1979: OK.

pithos/application.py
JasonLG1979: Reader is basically a generator. I don't see a way to easily get the size/length of the logs which is kinda stupid. I'll have to poop the logs in a list and then check the size of the list.

JasonLG1979: Never mind I guess I can cheat a little to avoid iterating over the logs twice.

pithos/application.py
```diff
@@ -85,26 +87,55 @@ def do_command_line(self, command_line):
...
+
+            _PRIORITY_TO_LEVEL = {
+                journal.LOG_DEBUG: 'DEBUG',
+                journal.LOG_INFO: 'INFO',
+                journal.LOG_WARNING: 'WARNING',
+                journal.LOG_ERR: 'ERROR',
+                journal.LOG_CRIT: 'CRTICIAL',
+                journal.LOG_ALERT: 'ALERT',
+            }
+
```
TingPing: __[flake8]__

pithos/application.py
```diff
@@ -85,26 +87,55 @@ def do_command_line(self, command_line):
...
+            _PRIORITY_TO_LEVEL = {
+                journal.LOG_DEBUG: 'DEBUG',
+                journal.LOG_INFO: 'INFO',
+                journal.LOG_WARNING: 'WARNING',
+                journal.LOG_ERR: 'ERROR',
+                journal.LOG_CRIT: 'CRTICIAL',
+                journal.LOG_ALERT: 'ALERT',
+            }
+            
+            got_logs = False
```
TingPing: __[flake8]__
* [W291] trailing whitespace

pithos/application.py
```diff
@@ -85,26 +87,55 @@ def do_command_line(self, command_line):
...
+                line = entry['CODE_LINE']
+                function = entry['CODE_FUNC']
+                module = basename(entry['CODE_FILE'])[:-3]
+                message = entry['MESSAGE']
+                log_line = '{} - {}:{}:{} - {}'.format(level, module, function, line, message)
+                self._print(command_line, log_line)
+
+            if not got_logs:
+                self._print(command_line, 'No Pithos logs present for this boot.')
+
```
TingPing: __[flake8]__

pithos/plugins/journald_logging.py
```diff
@@ -0,0 +1,144 @@
...
+
+        def on_dialog_response(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.hide()
+                self.emit('logging-changed', active_id)
+            dialog.destroy()
+
+        message = (_('The debug logging level is not '
+                   'recommended unless you are actually debugging an issue, '
+                   'as it generates very large logs.\n\nAre you sure you want to set logging to debug?'))
```
TingPing: __[flake8]__
* [E128] continuation line under-indented for visual indent

pithos/plugins/journald_logging.py
```diff
@@ -0,0 +1,146 @@
...
+        def on_dialog_response(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.hide()
+                self.emit('logging-changed', active_id)
+            dialog.destroy()
+
+        message = (
+            _('The debug logging level is not '
+            'recommended unless you are actually debugging an issue, '
+            'as it generates very large logs.\n\nAre you sure you want to set logging to debug?'),
```
TingPing: __[flake8]__
* [E128] continuation line under-indented for visual indent

JasonLG1979:
Waiting on you to do whatever you need to do @TingPing I'm pretty sure I changed everything you wanted?

pithos/application.py
```diff
@@ -85,26 +87,55 @@ def do_command_line(self, command_line):
...
         logging.root.handlers = []
 
+        # Show the Pithos log since last reboot and exit
+        if options.contains('last-logs'):
+            try:
+                from systemd import journal
+                from os.path import basename
+            except ImportError:
+                self._print(command_line, _('Systemd Python module not found'))
+                return 0
```
TingPing: Return non-zero on error.

JasonLG1979: What then?

TingPing: I don't understand the question, On an error processes just exit with a non-zero number, any will do.

JasonLG1979: I wasn't sure if it mattered, sorrrrry,lol!!!

JasonLG1979:
That work?

TingPing:
```python
Traceback (most recent call last):
  File "/app/lib/python3.4/site-packages/pithos/application.py", line 157, in do_command_line
    self.do_activate()
  File "/app/lib/python3.4/site-packages/pithos/application.py", line 170, in do_activate
    self.window = PithosWindow(self, self.test_mode)
  File "/app/lib/python3.4/site-packages/pithos/pithos.py", line 256, in __init__
    load_plugins(self)
  File "/app/lib/python3.4/site-packages/pithos/plugin.py", line 107, in load_plugins
    plugin = plugins[name] = load_plugin(name, window)
  File "/app/lib/python3.4/site-packages/pithos/plugin.py", line 70, in load_plugin
    module = __import__('pithos.plugins.' + name)
  File "/app/lib/python3.4/site-packages/pithos/plugins/journald_logging.py", line 76
    **kwargs,
            ^
SyntaxError: invalid syntax
```

PR #418, opened 2017-02-11
More complete PULSE_PROPs

JasonLG1979:
Useful for applications that interact with PulseAudio like https://github.com/masmu/pulseaudio-dlna

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -19,6 +19,7 @@ import json
 import logging
 import math
 import re
+import os
 import sys
 import time
 import tempfile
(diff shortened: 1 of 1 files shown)


TingPing:
Are these properties documented anywhere?

JasonLG1979:
http://0pointer.de/lennart/projects/pulseaudio/doxygen/proplist_8h.html

JasonLG1979:
> Please include that link in the commit message.

OK

> Also 'AudioStream' seems a bit vague?

I know I didn't know what to put there. That's what Firefox uses for the name?

JasonLG1979:
I can do something like 'Pandora Radio Stream' ?

TingPing:
Based upon the docs it seems that `Artist - Song` or such would be acceptable.

JasonLG1979:
So skip name then?

TingPing:
Hmm? I'm saying `name` is basically just a user friendly name of the current song.

JasonLG1979:
Artist and song are already covered with artist and title

JasonLG1979:
Name would be the name of the stream. Something in my mind would not change. A general description of the stream.

TingPing:
The example given is `e.g. "Guns'N'Roses: Civil War"`. I read that as a user readable name of the song.

JasonLG1979:
Didn't see that ok.

JasonLG1979:
I'll match the example then.

JasonLG1979:
Done

PR #422, opened 2017-02-11
Symbolic icon

JasonLG1979:
Adds a symbolic icon for display in the panel among other things?
screenshot from 2017-02-11 02-35-46


JasonLG1979:
Last One I swear. Deleting those unneeded icons got me thinking that I hadn't done a symbolic icon yet...

TingPing:
Think we should use symbolic icons for the tray too? It probably fits in better in gnome and kde at least.

JasonLG1979:
> Think we should use symbolic icons for the tray too? It probably fits in better in gnome and kde at least.


Then we have to sniff the DE, because we don't want all DE's to get a symbolic icon. And sniffing the DE is spotty at best especially with flatpak. Unless you have a bulletproof way to detect the DE? I'd rather just discourage use of that plugin all together. If it were up to me we'd get rid of it. It's been nothing but a PITA for functionality that 99.999% of users can and do get from our MPRIS plugin.

PR #424, opened 2017-02-12
Don't double tap position mpris

JasonLG1979:
There is a small chance that the 2nd call will return None and cause an exception. This will fix that by caching the position.


JasonLG1979:
OK, How about we release a 1.3 beta and push it to a staging PPA and invite people to test things out so we can do a solid release?

TingPing:
In my experience betas are largely useless, especially for a project as small as pithos. The few that would use betas can already use the git versions.

JasonLG1979:
Point taken

PR #428, opened 2017-02-17
fix the plugin prefs button tooltip

JasonLG1979:
It always says "This plugin either must be enabled or does not support preferences." This fixes that by show 'Click to change the *plugin name* plugin's settings'


JasonLG1979:
There what do you think of that?

JasonLG1979:
Really though all the plugins should have a "name" var so that their name can be translated. I can do that and go though the code base real quick and make all translatable strings well translatable,lol ;P

JasonLG1979:
@TingPing bump...

PR #431, opened 2017-02-25
Add MPRIS TrackList Interface Support

JasonLG1979:
Added support for the Tracklist interface, added all optional MPRISv2.2 Properties for the sake of completeness, did a little bit of clean up in the imports to shut up flake8, changed a couple arg names to better match spec when introspecting with apps like d-feet and updated the doc strings.


JasonLG1979:
In conjunction with this https://github.com/eonpatapon/gnome-shell-extensions-mediaplayer/pull/320 we will have a pretty sweet setup!!!

JasonLG1979:
I opted not to remove tracks that were banned or set tired because if/when we extend the interface with Pithos Ratings I'll want to keep them in the tracklist to be able to change their ratings.

JasonLG1979:
And also as I've said with this we will be the 1st player to have a complete MPRIS implementation. Well the 1st to have one that actually works. A couple players *claim* to have a complete implementation but they are broken in various ways.

pithos/pandora/pandora.py
```diff
@@ -586,3 +586,4 @@ def __init__(self, resultType, d):
         elif resultType == 'genre':
             self.stationName = d['stationName']
 
+
```
TingPing: __[flake8]__
* [W391] blank line at end of file

JasonLG1979:
I just merged support for TrackList interface in the media extension so we now have a client that can take advantage of this.

JasonLG1979:
I also threw in a few fixes and now the active player's symbolic icon is displayed in the panel instead of a generic audio icon if available(or optionally the current cover ofc) same goes for the player's menu icon(except for no cover art there. that would just be weird...) Anyway after this all I need to do is extend our mpris interface to support ratings and do the corresponding work to support Pithos ratings in the extension and the extension can effectively replace the notification icon plugin for GNOME users  workspace 1_004

JasonLG1979:
@TingPing bump...

PR #441, opened 2017-04-08
SecretService as a proper class

JasonLG1979:
This makes SecretService a proper class and seems to also solve the issue mentioned in https://github.com/pithos/pithos/pull/390 by caching the password.

### DIFF
--- a/pithos/util.py
+++ b/pithos/util.py
@@ -24,76 +24,101 @@ from gi.repository import (
     Gtk
 )
 
-_ACCOUNT_SCHEMA = Secret.Schema.new('io.github.Pithos.Account', Secret.SchemaFlags.NONE,
-                                    {"email": Secret.SchemaAttributeType.STRING})
 
-_current_collection = Secret.COLLECTION_DEFAULT
-
-
-# TODO: Async
-def unlock_keyring():
-    global _current_collection
-
-    service = Secret.Service.get_sync(
-        Secret.ServiceFlags.NONE,
-        None,
+class _SecretService:
+    _cached_password = ''
+    _current_collection = Secret.COLLECTION_DEFAULT
+    _account_schema = Secret.Schema.new(
+        'io.github.Pithos.Account',
+        Secret.SchemaFlags.NONE,
+        {'email': Secret.SchemaAttributeType.STRING},
     )
 
-    default_collection = Secret.Collection.for_alias_sync(
-        service,
-        Secret.COLLECTION_DEFAULT,
-        Secret.CollectionFlags.NONE,
-        None,
-    )
-
-    if default_collection is None:
-        logging.warning(
-            'Could not get the default Secret Collection.\n'
-            'Attempting to use the session Collection.'
-        )
-        _current_collection = Secret.COLLECTION_SESSION
-        return
-
-    if not default_collection.get_locked():
-        logging.debug('The default keyring is unlocked.')
-    else:
-        num_items, unlocked = service.unlock_sync(
-            [default_collection],
+    def unlock_keyring(self):
+        service = Secret.Service.get_sync(
+            Secret.ServiceFlags.NONE,
             None,
         )
 
-        if not num_items or default_collection not in unlocked:
-            _current_collection = Secret.COLLECTION_SESSION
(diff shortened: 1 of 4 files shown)


pithos/util.py
```diff
@@ -24,50 +24,94 @@
     Gtk
 )
 
-_ACCOUNT_SCHEMA = Secret.Schema.new('io.github.Pithos.Account', Secret.SchemaFlags.NONE,
-                                    {"email": Secret.SchemaAttributeType.STRING})
+class _SecretService:
```
TingPing: __[flake8]__

pithos/util.py
```diff
@@ -24,50 +24,94 @@
...
-        if not num_items or default_collection not in unlocked:
-            _current_collection = Secret.COLLECTION_SESSION
-            logging.debug('The default keyring is locked. Using session collection.')
-        else:
-            logging.debug('The default keyring was unlocked.')
+        self._cached_password = password
+
+        return True
 
+SecretService = _SecretService()
```
TingPing: __[flake8]__
* [E305] expected 2 blank lines after class or function definition, found 1

pithos/util.py
```diff
@@ -24,49 +24,95 @@
...
-        if not num_items or default_collection not in unlocked:
-            _current_collection = Secret.COLLECTION_SESSION
-            logging.debug('The default keyring is locked. Using session collection.')
-        else:
-            logging.debug('The default keyring was unlocked.')
+        self._cached_password = password
+
+        return True
+
+SecretService = _SecretService()
```
TingPing: __[flake8]__
* [E305] expected 2 blank lines after class or function definition, found 1

TingPing:
I'm not sure this makes things much cleaner.

JasonLG1979:
It encapsulates everything. It's better that having a bunch of related free functions just floating there isn't it? and as I said caching the password seems to solve the problem in 390. I think we are running into a situation where we're trying to get the password to fast after we set it. Caching it means we don't have to ask the secret service for it as much.

JasonLG1979:
From a dogmatic python standpoint aside from caching the password we're trading globals for effectively a singleton. Both are viewed as bad but in this case a singleton is actually what we want.

pithos/util.py
```diff
@@ -24,76 +24,101 @@
...
+            {'email': email},
+            self._current_collection,
+            'Pandora Account',
+            password,
+            None,
+        )
 
-    if not password:
-        return _clear_account_password(email)
+        if password_store_successful:
```
TingPing: __[flake8]__
* [W291] trailing whitespace

pithos/util.py
```diff
@@ -24,76 +24,101 @@
...
+            None,
+        )
 
-    if not password:
-        return _clear_account_password(email)
+        if password_store_successful: 
+            self._cached_password = password
+        else:
+            self._cached_password = ''
+            logging.warning('Failed to store password')
```
TingPing: __[flake8]__
* [W291] trailing whitespace

JasonLG1979:
@TingPing, This also seems to solve the problem where on 1st if I login to my Pandora One account Pithos quits. For the same reason as I mentioned above I'm guessing. I think we're asking for the password to fast after we set it.

JasonLG1979:
I'd really like to see this and the ratings extension merged. I'm still playing with the Lasf.fm PR. We don't use much of pylast and the Last.fm api is pretty simple. I was toying with the idea of just ditching pylast and doing my own thing with Last.fm.

TingPing:
We can't "cache" the password, its an external service that can change beyond our control and without notifying us.

JasonLG1979:
> its an external service that can change beyond our control and without notifying us.

I'm not sure why anything would change the password set for and by Pithos besides Pithos?

JasonLG1979:
The "external service" is apparently to slow. The alternative is to go async. But that's really tricky.

JasonLG1979:
If we go async it gets worse. You set the password async and then ask for it async. There's a good chance they won't return in the right order.

TingPing:
I believe the order of DBus messages is retained. We should being going async anyway.

JasonLG1979:
OK I already have a branch where I started it.

PR #443, opened 2017-04-09
Last fm integration improvements

JasonLG1979:
Use a headerbar.

Better logging.

The auth dialog now pops up when the plugin is enabled but Pithos is not authorized with Last.fm.
The plugin is also disabled if the dialog is closed and Pithos is not authorized with Last.fm.
This is to avoid confusion as to if the plugin is actually really enabled.

Use pylast.LastFMNetwork preferably instead of pylast.get_lastfm_network.

Show a dialog that tells the user the account url that the Last.fm plugin is authorized for and lets them decide if they want to continue to use that account on a Pandora login change.
screenshot from 2017-04-09 07-37-58
screenshot from 2017-04-10 10-38-27
screenshot from 2017-04-10 10-38-08

### DIFF
--- a/pithos/plugins/lastfm.py
+++ b/pithos/plugins/lastfm.py
@@ -44,6 +44,7 @@ class LastfmPlugin(PithosPlugin):
         self.worker = GObjectWorker()
         self.preferences_dialog = LastFmAuth(self.pylast, self.settings)
         self.preferences_dialog.connect('lastfm-authorized', self.on_lastfm_authorized)
+        self.window.prefs_dlg.connect('login-changed', self._show_dialog)
 
     def on_enable(self):
         if self.settings['data']:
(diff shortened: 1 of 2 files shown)


pithos/plugins/lastfm.py
```diff
@@ -24,153 +25,273 @@
...
+                self.preferences_dialog.auth_state = self.preferences_dialog.AuthState.NOT_AUTHORIZED
+                self.settings.reset('enabled')
+                self.settings.reset('data')
+                self.preferences_dialog.button.set_sensitive(True)
+                self.preferences_dialog.set_widget_text()
+            dialog.destroy()
+
+        sub_text = _('Would you like to continue using this account?')
+        sub_text = '{}\n{}'.format(sub_text, user_profile_url)
+        secondary_text = _('If not you will need to re-enable the Last.fm Plugin if you wish to use a diffrent account.')
```
TingPing: __[flake8]__
* [E501] line too long (121 > 120 characters)

pithos/plugins/lastfm.py
```diff
@@ -24,153 +25,303 @@
...
-        widget.hide()
-        return True # Don't delete window
+            text = _('The Last.fm Plugin is Disabled')
+            secondary_text = _('But Pithos is still authorized with this Last.fm account:')
+            trinary_text = _('Would you like to deauthorize it?')
+
+        if self.window.prefs_dlg.get_visible():
+            parent = self.window.prefs_dlg
+        else:
+            parent = self.window
```
TingPing: __[flake8]__
* [W291] trailing whitespace

PR #459, opened 2017-04-27
Detect the Thumbprint Radio Station

JasonLG1979:
This will make it so we treat the Thumbprint Radio similar to the QuickMix. It will always be right after the QuickMix in the stations popover and will not show up in the station dialog since it can't be renamed. It will also always be the 2nd station in MPRIS.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -648,7 +648,11 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.stations_popover.clear()
         self.current_station = None
         selected = None
-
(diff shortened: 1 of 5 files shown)


pithos/pithos.py
```diff
@@ -648,7 +648,12 @@ def process_stations(self, *ignore):
         self.stations_popover.clear()
         self.current_station = None
         selected = None
-
+        # Make sure that the Thumprint Radio Station is always 2nd.
+        for i, s in enumerate(self.pandora.stations):
+            if s.isThumbprint:
+                self.pandora.stations.insert(1, self.pandora.stations.pop(i))
+                break
+
```

PR #462, opened 2017-05-02
Auto Pandora One

JasonLG1979:
This will automatically set the client free or Pandora One. It removes the need for the Pandora One Checkbox.

### DIFF
--- a/pithos/PreferencesPithosDialog.py
+++ b/pithos/PreferencesPithosDialog.py
@@ -108,7 +108,6 @@ class PreferencesPithosDialog(Gtk.Dialog):
     proxy_entry = GtkTemplate.Child()
     control_proxy_entry = GtkTemplate.Child()
     control_proxy_pac_entry = GtkTemplate.Child()
-    pandora_one_checkbutton = GtkTemplate.Child()
     explicit_content_filter_checkbutton = GtkTemplate.Child()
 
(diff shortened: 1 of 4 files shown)


JasonLG1979:
If you don't have a Pandora One account @TingPing I'll email you my creds for my main account so you can test this.

JasonLG1979:
This also has the benefit that it doesn't require any sort of migrating of settings.

TingPing:
I don't but I'll go on good faith you tested it throughly enough as its a fairly small patch.

pithos/PreferencesPithosDialog.py
```diff
@@ -213,11 +210,10 @@ def cb(success):
...
             email = self.email_entry.get_text()
             password = self.password_entry.get_text()
-            pandora_one = self.pandora_one_checkbutton.get_active()
 
-            if self.last_email != email or self.last_password != password or self.last_pandora_one != pandora_one:
+            if self.last_email != email or self.last_password != password:
                 SecretService.set_account_password(self.last_email, email, password, cb)
             else:
-                self.settings.revert()
+                self.settings.apply()
```
TingPing: What was this change for?

JasonLG1979: That's a fix from the last bit. Otherwise just changing the audio quality setting has no effect.

PR #464, opened 2017-05-03
Common SESSION Bus

JasonLG1979:
This goes a long way to making plugins that can be loaded async, async.

### DIFF
--- a/pithos/plugin.py
+++ b/pithos/plugin.py
@@ -16,7 +16,10 @@
 import logging
 import glob
 import os
-from gi.repository import Gio
+from gi.repository import (
+    GLib,
+    Gio
+)
 
 
 class PithosPlugin:
@@ -24,9 +27,10 @@ class PithosPlugin:
     preference = None
     description = ""
 
-    def __init__(self, name, window):
+    def __init__(self, name, window, bus):
         self.name = name
         self.window = window
+        self.bus = bus
         self.preferences_dialog = None
         self.prepared = False
         self.enabled = False
@@ -65,7 +69,7 @@ class ErrorPlugin(PithosPlugin):
         self.enabled = False
 
 
-def load_plugin(name, window):
+def load_plugin(name, window, bus):
     try:
         module = __import__('pithos.plugins.' + name)
         module = getattr(module.plugins, name)
@@ -81,7 +85,7 @@ def load_plugin(name, window):
     else:
         return ErrorPlugin(name, "Could not find module class")
 
-    return plugin_class(name, window)
+    return plugin_class(name, window, bus)
 
 
 def _maybe_migrate_setting(new_setting, name):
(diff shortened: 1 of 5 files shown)


pithos/plugin.py
```diff
@@ -95,29 +96,46 @@ def _maybe_migrate_setting(new_setting, name):
...
-            plugin.settings = settings.get_child(settings_name)
-            _maybe_migrate_setting(plugin.settings, name)
-        else:
-            # Out of tree plugin
-            plugin.settings = Gio.Settings.new_with_path('io.github.Pithos.plugin',
+    def on_got_bus(source, result, userdata):
+        try:
+            bus = Gio.bus_get_finish(result)
+            logging.info('Got session bus')
+        except Glib.Error as e:
```
TingPing: __[flake8]__
* [F821] undefined name 'Glib'

pithos/plugin.py
```diff
@@ -16,17 +16,21 @@
 import logging
 import glob
 import os
-from gi.repository import Gio
+from gi.repository import (
```
TingPing: __[flake8]__
* [F401] 'gi.repository.GLib' imported but unused

pithos/plugin.py
```diff
@@ -95,29 +99,46 @@ def _maybe_migrate_setting(new_setting, name):
...
-            plugin.settings = settings.get_child(settings_name)
-            _maybe_migrate_setting(plugin.settings, name)
-        else:
-            # Out of tree plugin
-            plugin.settings = Gio.Settings.new_with_path('io.github.Pithos.plugin',
+    def on_got_bus(source, result, userdata):
+        try:
+            bus = Gio.bus_get_finish(result)
+            logging.info('Got session bus')
+        except Glib.Error as e:
```
TingPing: __[flake8]__
* [F821] undefined name 'Glib'

TingPing:
Screensaver should handle in prepare not enable. Notify doesn't use this?

Also better commit titles man: `plugin: Share session bus between plugins`

TingPing:
Mediakeys could also better handle an error message in the case that keybinder and dbus don't work.

JasonLG1979:
> Screensaver should handle in prepare not enable.

OK.

> Notify doesn't use this?

It does but not in that way. It gets the session bus when it inits. For GioNotify it'll be a little more tricky we would need to handle the error in a callback.

> Mediakeys could also better handle an error message in the case that keybinder and dbus don't work.

Ok.

JasonLG1979:
> Mediakeys could also better handle an error message in the case that keybinder and dbus don't work.

```python
return 'Keybinder requires X11'
```
and
```python
return 'Keybinder not found'
```

Could be changed to:

```python
return 'DBus binding failed and Keybinder requires X11'
```
and
```python
return 'DBus binding failed and Keybinder not found'
```

JasonLG1979:
Something like that?

PR #466, opened 2017-05-04
Async plugins

JasonLG1979:


pithos/plugins/dbus_util/GioNotify.py
```diff
@@ -70,49 +70,57 @@ def __init__(self, **kwargs):
...
+            else:
+                self.call(
+                    'GetServerInformation',
+                    None,
+                    Gio.DBusCallFlags.NONE,
+                    -1,
+                    None,
+                    on_GetServerInformation_finish,
+                    None,
+               )
```
TingPing: __[flake8]__
* [E122] continuation line missing indentation or outdented

JasonLG1979:
I think that about covers it @TingPing can you think of anywhere else we can go async?

JasonLG1979:
Do you want me to remove the stuff in GioNotify we don't use anymore?

JasonLG1979:
> Do you want me to remove the stuff in GioNotify we don't use anymore?

@TingPing? yea or nah?

TingPing:
I don't know what we don't use. I don't expect anybody else to use it so sure.

JasonLG1979:
OK

JasonLG1979:
This too!!!

pithos/plugins/notification_icon.py
```diff
@@ -13,56 +13,63 @@
...
+        try:
+            gi.require_version('AppIndicator3', '0.1')
+            from gi.repository import AppIndicator3 as AppIndicator
+            self.indicator_capable = True
+        except (ImportError, ValueError):
+            self.indicator_capable = False
+
+        if self.indicator_capable:
+            try:
+                self.ind = AppIndicator.Indicator.new(
```
TingPing: Can this raise an exception?

JasonLG1979: I'm not sure? I just wanted to cover our bases since the tray icon doesn't work with Wayland. I'm don't even know how to test AppIndicator against Wayland? I suppose I can install  AppIndicator in my Ubuntu GNOME VM and try it out in the Wayland session with the AppIndicator Shell extension?

TingPing: > I suppose I can install AppIndicator in my Ubuntu GNOME VM and try it out in the Wayland session with the AppIndicator Shell extension?

Probably.

> I'm not sure?

It indeed doesn't ever fail: http://bazaar.launchpad.net/~indicator-applet-developers/libappindicator/trunk.16.10/view/head:/src/app-indicator.c#L1766

JasonLG1979: You're right it doesn't ever fail which is bad. Without that extension if a user has AppIndicator3 installed they get no tray icon and when they try to close Pithos they loose the window.

JasonLG1979: I really hate that plugin...

JasonLG1979: We're going to have to check for that extension if we're on GNOME...

JasonLG1979: Btw, I did as I said installed that extension on Ubuntu GNOME and appindicator works fine in Wayland. I can't say the same for the extension. Sometimes it shows up in GNOME tweak sometimes not...

TingPing: Yes we have to detect if on Wayland and DE doesn't support AppIndicator somehow. Maybe we can check if some dbus name is owned?

TingPing: http://bazaar.launchpad.net/~indicator-applet-developers/libappindicator/trunk.16.10/view/head:/src/dbus-shared.h

JasonLG1979: Working around the inconsistent behavior is a separate PR though.

JasonLG1979: > Yes we have to detect if on Wayland and DE doesn't support AppIndicator somehow. Maybe we can check if some dbus name is owned?

I'll take a look into it after you merge this PR. As I said that's a separate PR unless you just want me to add a commit for it in this PR?

JasonLG1979: Checking for indicator applet interface would work I'd imagine for *buntu's that use real indicator applet. But the extension doesn't appear to be owning any names from what I can see in D-feet. It looks like it's just using the org.kde.StatusNotifierWatcher to watch for stuff to put in the panel.

JasonLG1979: What we can do if appindicator is installed it 1st check for the real indicator interface and then if we don't find it check if the extension is installed and enabled in gsettings.

JasonLG1979:
> Yes we have to detect if on Wayland and DE doesn't support AppIndicator somehow. Maybe we can check if some dbus name is owned?

Done. Tested on GNOME with and without the extension. On KDE with and without AppIndicator3 installed and on *buntu's with the Indicator applet.

Seems to work as expected.

JasonLG1979:
Ok @TingPing so a summary of the changes:

1. Plugins that can be loaded async are.

2. Removed unused code from GioNotify.

3. The notification_icon plugin is now much smarter about loading. In GNOME Shell in Wayland it will fail gracefully as before. In GNOME Shell with AppIndicator3 and the appindicatorsupport GNOME Shell extension users get a appindicator tray icon even in Wayland. Buntu's with the indicator applet get a appindicator tray icon. KDE with AppIndicator3 gets a appindicator tray icon. KDE without AppIndicator3 fails gracefully. Everyone else gets  Gtk.StatusIcon.

4. The mediakeys plugin now fails gracefully in on_enable.

5. Check to see if the bus name is owned instead of calling methods and seeing if they fail in GioNotify and the mediakeys plugin.

pithos/plugins/notification_icon.py
```diff
@@ -12,57 +12,142 @@
...
-        self.delete_callback_handle = self.window.connect("delete-event", self._toggle_visible)
-        self.state_callback_handle = self.window.connect("play-state-changed", self.play_state_changed)
-        self.song_callback_handle = self.window.connect("song-changed", self.song_changed)
+        def has_appindicator_shell_ext():
+            schema = Gio.SettingsSchemaSource.get_default()
+            valid_schema = schema.lookup('org.gnome.shell', True)
+            if valid_schema:
+                gnome_shell = Gio.Settings.new('org.gnome.shell')
+                enabled_extensions = gnome_shell.get_value('enabled-extensions')
+                if 'appindicatorsupport@rgcjonas.gmail.com' in enabled_extensions:
```
TingPing: I don't think this is the right way, we should be checking for the kde bus name it owns.

pithos/plugins/notification_icon.py
```diff
@@ -12,57 +12,142 @@
...
+            if not type(display).__name__.endswith('X11Display'):
+                logging.debug('Notification icon requires X11 or AppIndicator Support')
+                self.prepare_complete(error='Notification icon requires X11 or AppIndicator Support')
+            else:
+                self.prepare_complete()
+
+        def init_appindicator():
+            self.ind = self.appindicator.Indicator.new(
+                "pithos-tray-icon",
+                "pithos-tray-icon",
```
TingPing: This needs rebased on master.

pithos/plugins/notification_icon.py
```diff
@@ -12,57 +12,142 @@
...
+                else:
+                    check_for_x()
+
+        def on_indicator_finish(source, result, data):
+            try:
+                indicator_service = Gio.DBusProxy.new_finish(result)
+            except GLib.Error as e:
+                logging.debug(e)
+                check_for_x()
+            else:
```
TingPing: Could you sprinkle a few logging.debug's through-out here, it is sort of a complex startup trying a few different methods and it would be useful.

pithos/plugins/notification_icon.py
JasonLG1979: It doesn't own any names it just listens on the bus.

TingPing: I installed it as a test and it indeed owns that bus name.

JasonLG1979: Which one?

TingPing: `org.kde.StatusNotifierWatcher`

JasonLG1979: Yep, sure does. That simplifies things KDE and the Ubuntu indicator applet also own that same name. So I can just check to see if that name is owned.

pithos/plugins/notification_icon.py
```diff
@@ -12,55 +12,121 @@
...
+        else:
+            Gio.DBusProxy.new(
+                self.bus,
+                Gio.DBusProxyFlags.DO_NOT_LOAD_PROPERTIES,
+                None,
+                'org.kde.StatusNotifierWatcher',
+                '/StatusNotifierWatcher',
+                'org.kde.StatusNotifierWatcher',
+                None,
+                n_StatusNotifierWatcher_finish,
```
* [F821] undefined name 'n_StatusNotifierWatcher_finish'

pithos/plugins/notification_icon.py
JasonLG1979: Done.

pithos/plugins/notification_icon.py
```diff
@@ -112,6 +79,18 @@ def check_for_plasma():
...
+            try:
+                StatusNotifierWatcher = Gio.DBusProxy.new_finish(result)
+            except GLib.Error as e:
+                logging.debug(e)
+                check_for_x()
+            else:
+                if StatusNotifierWatcher.get_name_owner():
+                    init_appindicator()
+                else:
+                    check_for_plasma()
```
TingPing: Why do we still need to check for plasma, do they not expose the same `org.kde.StatusNotifierWatcher`?

JasonLG1979: You're right. Right there we shouldn't be checking for Plasma. My bad it's hard sometime to keep track of the flow with this much going on,lol!!!

JasonLG1979: We still need ot check for Plasma though if AppIndicator3 is not installed though. It's possible to have AppIndicator not installed, and that bus still owned on Plasma. Plasma doesn't support Gtk.StatusIcon.

TingPing: As I understand it Plasma reverted that stance and does support them again?

TingPing: Also this might be beyond reasonable, but appindicator is trash, I just built it for the flatpak. Might make more sense to rewrite it in a few lines of python...

JasonLG1979: > As I understand it Plasma reverted that stance and does support them again?

It shows up on Kubuntu 17.04 but is non-functional. It just sits there, you can't interact with it.

JasonLG1979: > Also this might be beyond reasonable, but appindicator is trash, I just built it for the flatpak. Might make more sense to rewrite it in a few lines of python...

I'll look at the source and see what I can come up with.

JasonLG1979: Have I mentioned how much I hate that plugin?

JasonLG1979: >Might make more sense to rewrite it in a few lines of python...
I haven't looked at the source yet but that would basically involve creating a DBus interface with your DBusServiceObject correct?

TingPing: Yea, at a glance maybe it is more work than its worth.

JasonLG1979: > Yea, at a glance maybe it is more work than its worth.

Maybe but here is a starting point I think. https://freedesktop.org/wiki/Specifications/StatusNotifierItem/

If I can decipher MPRIS I can prop figure this out. Just might take a while.

TingPing: Probably best to implement the KDE spec not the Unity spec: https://github.com/jjk-jacky/statusnotifier

TingPing: Oh beat me to it =)

TingPing: Ugh the KDE spec seems decent but it depends upon the dbusmenu canonical spec which looks awful still.

JasonLG1979: > Ugh the KDE spec seems decent but it depends upon the dbusmenu canonical spec which looks awful still.

Yet another DBus interface we'd have to create I'm guessing? Piss on it. We've already put more work into this plugin then it deserves.

JasonLG1979: So to summarize if we wanted to re-implement AppIndicator3 in python we'd need to create 3 interfaces:

StatusNotifierHost
StatusNotifierItem
Dbusmenu

Basically aside from the Buntu's which already have AppIndicator3 available the only DE that it would actually work on is KDE Plasma right? Reading the docs for the spec it's a little suspect. For instance all the icon props have a redundant *Pixmap prop. It says that all props are mandatory yet it also says that it's not possible to set the normal icon prop and the *Pixmap prop at the same time.

JasonLG1979: I'm not sure why it all couldn't be one interface it seems really stupid that there are 3 interfaces since all are required to get the desired end result StatusNotifierHost seems especially useless...

JasonLG1979: Why would you have an interface with no properties, methods or signals?

JasonLG1979: I still may give it a shot for the challenge and novelty of it but I really don't see the actually usefulness of it. I'm fine not supporting tray icons in KDE Plasma outside of the Buntu's for now.

Your thoughts @TingPing?

JasonLG1979:
@TingPing Do you want me to squash these commits and basically make https://github.com/pithos/pithos/pull/466#issuecomment-299702670 my commit message?

TingPing:
No

JasonLG1979:
> No

So maybe merge it like this?

TingPing:
I merged everything except the notification icon changes as I want to take more time to review/test that. If you don't mind make a new PR over just those changes.

JasonLG1979:
> I merged everything except the notification icon changes as I want to take more time to review/test that. If you don't mind make a new PR over just those changes.

No problem, can do.

PR #468, opened 2017-05-12
Check for AppIndicator3 support

JasonLG1979:
This checks for the StatusNotifierWatcher interface before we try to use AppIndicator3.(Which AppIndicator3 should do but does not.) If StatusNotifierWatcher does exsist assume we are in a DE that doesn't support Gtk.StatusIcon (Unity or KDE Plasma) if the StatusNotifierWatcher interface does not exsist check for the X server to make sure we can use Gtk.StatusIcon.

### DIFF
--- a/pithos/plugins/notification_icon.py
+++ b/pithos/plugins/notification_icon.py
@@ -12,28 +12,18 @@
 # You should have received a copy of the GNU General Public License along
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import logging
 import os
-import sys
-import gi
 
-from gi.repository import GObject, Gdk, Gtk
+from gi.repository import (
+    GObject,
+    GLib,
+    Gtk,
+    Gdk,
+    Gio
+)
 
 from pithos.plugin import PithosPlugin
-
-# Use appindicator if installed
-try:
-    gi.require_version('AppIndicator3', '0.1')
-    from gi.repository import AppIndicator3 as AppIndicator
-    indicator_capable = True
-except (ImportError, ValueError):
-    indicator_capable = False
-
-
-def backend_is_supported(window):
-    if sys.platform in ('win32', 'darwin'):
-        return True
-    display = window.props.screen.get_display()
-    return type(display).__name__.endswith('X11Display')
 
 
 def get_local_icon_path():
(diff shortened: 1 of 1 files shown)


pithos/plugins/notification_icon.py
```diff
@@ -12,57 +12,120 @@
...
+                )
 
-    def on_enable(self):
-        self.delete_callback_handle = self.window.connect("delete-event", self._toggle_visible)
-        self.state_callback_handle = self.window.connect("play-state-changed", self.play_state_changed)
-        self.song_callback_handle = self.window.connect("song-changed", self.song_changed)
 
-        if indicator_capable:
-            self.ind.set_status(AppIndicator.IndicatorStatus.ACTIVE)
+    def on_enable(self):
```
* [E303] too many blank lines (2)

JasonLG1979:
Looking for a place to file a bug against AppIndicator3 it really should check for that interface and throw an error if it's not found since it's not functional without it... I hate launchpad...

pithos/plugins/notification_icon.py
```diff
@@ -12,57 +12,119 @@
...
+                logging.debug(e)
+                check_for_x()
+            else:
+                if plasmashell.get_name_owner():
+                    logging.debug('Notification icon requires AppIndicator on Plasma')
+                    self.prepare_complete(error='Notification icon requires AppIndicator on Plasma')
+                else:
+                    check_for_x()
+
+        def check_for_plasma():
```
TingPing: I forget if I already asked, but does Plasma not own `org.kde.StatusNotifierWatcher` ?

pithos/plugins/notification_icon.py
```diff
@@ -83,31 +145,32 @@ def button(text, action, checked=False):
...
             # On middle-click
             self.ind.set_secondary_activate_target(self.visible_check)
 
-        self.playpausebtn = button("Pause", self.window.playpause)[0]
-        button("Skip", self.window.next_song)
-        button("Love", (lambda *i: self.window.love_song()))
-        button("Ban", (lambda *i: self.window.ban_song()))
-        button("Tired", (lambda *i: self.window.tired_song()))
-        button("Quit", self.window.quit)
+        self.playpausebtn = button('Pause', self.window.playpause)[0]
```
TingPing: Lets avoid trivial syntax changes mixed in with real ones.

pithos/plugins/notification_icon.py
TingPing: Oh you are checking if the shell doesn't support XEmbed. How bout we assume if the shell provides `StatusNotifierWatcher` then it wants that over XEmbed and we always error out without AppIndicator?

JasonLG1979: > I forget if I already asked, but does Plasma not own org.kde.StatusNotifierWatcher ?

Yep. It does.

JasonLG1979: > Oh you are checking if the shell doesn't support XEmbed. How bout we assume if the shell provides StatusNotifierWatcher then it wants that over XEmbed and we always error out without AppIndicator?

Sure. Whatever you want, we can do that.

pithos/plugins/notification_icon.py
JasonLG1979: I just hate " 's...

pithos/plugins/notification_icon.py
JasonLG1979: Done. That also simplifies things by negating the need to check for the plasma shell. Which i think is what you were going for ;P

pithos/plugins/notification_icon.py
JasonLG1979: Single quotes everywhere just make everything more consistent. ATM Pithos has a mix. It bugs me...

JasonLG1979: > I forget if I already asked, but does Plasma not own org.kde.StatusNotifierWatcher ?

appindicator is basically a sh!t wrapper around a DBus service kinda like libnotify except ofc it's the creator not the consumer. KDE/Qt tray icons work in Unity with little to no modification without the aid of appindicator if I remember correctly.

pithos/plugins/notification_icon.py
JasonLG1979: > I forget if I already asked, but does Plasma not own org.kde.StatusNotifierWatcher ?

appindicator is basically a sh!t wrapper around a DBus service kinda like libnotify except ofc it's the creator not the consumer. KDE/Qt tray icons work in Unity with little to no modification without the aid of appindicator if I remember correctly.

TingPing: Right but as mentioned in the other thread we can't drop it fully because we would still depend upon `dbusmenu` probably.

pithos/plugins/notification_icon.py
```diff
@@ -117,20 +153,20 @@ def set_active(active):
         self.menu = menu
 
     def play_state_changed(self, window, playing):
-        """ play or pause and rotate the text """
+        ''' play or pause and rotate the text '''
```
TingPing: Did you actually go back and change *more* quotes or am I losing my mind?

JasonLG1979: No I changed them all at once the 1st time around and haven't changed them back yet,lol!!!

pithos/plugins/notification_icon.py
JasonLG1979: >Right but as mentioned in the other thread we can't drop it fully because we would still depend upon dbusmenu probably.

I'm not saying we should drop it right now. I'm saying that theoretically in the future we could if we implemented all the required DBus interfaces there's like 3 or 4 we'd have to do. It's stupid...

JasonLG1979: dbusmenu is just another DBus interface.

TingPing:
While on the topic of appindicator we should figure out better icons to use on gnome-shell/kde. The symbolic one should work for gs but what about kde?

JasonLG1979:
Why don't we just make that a preference? It would be much easier than a bunch of convoluted logic trying to sniff the DE.

TingPing:
@JasonLG1979 I think I covered all of the possible edge cases, let me know what you think ^

JasonLG1979:
That works. But I still like the idea of following the pattern of the other plugin and doing the imports that may fail in on_prepare. Functionally I don't think it matters, it's just a consistency thing.

JasonLG1979:
Do you want me to make a pref for choosing between full color and symbolic icon and if so which do you want me to default to?

TingPing:
> I tested the symbolic icon on gs and it was inconsistent depending on the theme it didn't always get the color of the icon right.

Could that be considered a bug in gs? Did you name the icon -symbolic (don't know if that actually matters). How does KDE handle it?

If it is a must we can do an option, I just want to know if it can be avoided.

JasonLG1979:
> Could that be considered a bug in gs? Did you name the icon -symbolic (don't know if that actually matters). How does KDE handle it?
> 
> If it is a must we can do an option, I just want to know if it can be avoided.

I just substituted with the existing symbolic icon with the gtk tray icon in GNOME. I haven't tested on GNOME or KDE yet with appindicator. My concern is that "legacy" style DE's like Mate, Xfce and LXDE might not handle the symbolic icon very well.

JasonLG1979:
It's going to have to be an option. The icon doesn't change color in KDE. It works just fine in Ubuntu GNOME with the appindicator extension.

TingPing:
We can probably just always use the symbolic one for the Adwaita icon theme, dunno if you would be interested for making icons specifically for kde or such.

JasonLG1979:
> We can probably just always use the symbolic one for the Adwaita icon theme, dunno if you would be interested for making icons specifically for kde or such.

Symbolic icons also do not work in Unity. I don't know how KDE handles it's symbolic icons but there was a bug filed against the appindicator shell extension about symbolic icons. I think the extension might be doing something extra? I'm starting to think AppIndicator3 is not symbolic icon aware.

TingPing:
Yea AppIndicator doesn't care what the icon is its purely up to the server to decide how to load and display it.

JasonLG1979:
Looking around it appears that KDE handles symbolic icons differently than GNOME it would require making making new icons and it still wouldn't work with Unity. It's really not worth the trouble it's easier just to make a prefs setting that defaults to full color with an option for symbolic.

TingPing:
I would be ok carrying another icon just for KDE if it can look better there but thats on you if you want to do it.

PR #469, opened 2017-05-14
notification_icon plugin icon setting

JasonLG1979:
This will allow the user to choose between a full color or symbolic notification tray icon.

### DIFF
--- a/pithos/plugins/notification_icon.py
+++ b/pithos/plugins/notification_icon.py
@@ -69,6 +69,7 @@ class PithosNotificationIcon(PithosPlugin):
                 self.prepare_complete(error='Failed to find DBus service')
                 return
 
+            self.preferences_dialog = NotificationIconPluginPrefsDialog(self.window, self.settings)
             # This is an awful mess but I don't know a better way of organizing it
             if is_owned and indicator_capable:
                 self._create_appindicator()
@@ -92,12 +93,16 @@ class PithosNotificationIcon(PithosPlugin):
                 # No fancy tray here
                 self.prepare_complete()
 
+        if not self.settings['data']:
+            self.settings['data'] = 'io.github.Pithos-tray'
+
         logging.info('Checking if org.kde.StatusNotifierWatcher is owned')
         self.bus.call('org.freedesktop.DBus', '/', 'org.freedesktop.DBus',
                       'NameHasOwner', GLib.Variant('(s)', ('org.kde.StatusNotifierWatcher',)),
                       GLib.VariantType('(b)'), Gio.DBusCallFlags.NONE, -1, None, on_has_name_owner)
 
     def on_enable(self):
+        self.icon_changed_handler = self.preferences_dialog.connect('icon-changed', self._on_icon_changed)
         self.delete_callback_handle = self.window.connect("delete-event", self._toggle_visible)
         self.state_callback_handle = self.window.connect("play-state-changed", self.play_state_changed)
         self.song_callback_handle = self.window.connect("song-changed", self.song_changed)
@@ -105,7 +110,7 @@ class PithosNotificationIcon(PithosPlugin):
         if indicator_capable:
             self.ind.set_status(AppIndicator.IndicatorStatus.ACTIVE)
         else:
-            self.statusicon = Gtk.StatusIcon.new_from_icon_name('io.github.Pithos-tray')
+            self.statusicon = Gtk.StatusIcon.new_from_icon_name(self.settings['data'])
             self.statusicon.connect('activate', self._toggle_visible)
 
         self.build_context_menu()
@@ -118,7 +123,7 @@ class PithosNotificationIcon(PithosPlugin):
 
     def _create_appindicator(self):
         self.ind = AppIndicator.Indicator.new("io.github.Pithos-tray",
-                                              "io.github.Pithos-tray",
+                                              self.settings['data'],
                                               AppIndicator.IndicatorCategory.APPLICATION_STATUS)
         local_icon_path = get_local_icon_path()
         if local_icon_path:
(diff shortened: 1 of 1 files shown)


pithos/plugins/notification_icon.py
```diff
@@ -210,6 +222,91 @@ def on_disable(self):
...
+            return
+
+        def on_dialog_response(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.hide()
+                self.emit('icon-changed', active_id)
+            dialog.destroy()
+
+        message = (_(
+            'Symbolic Notification Icons currently do not function correctly on all Desktop Environments or Configurations'
```
* [E501] line too long (123 > 120 characters)

JasonLG1979:
This is kinda a fix but still a mess. It works perfectly in GNOME Shell with the default icon theme with either the gtk tray icon or the appindicator extension but at least one icon theme (Papirus) inherits from Ubuntu-mono for it's full color icons so no matter the panel color you get a light color icon. I also can't test if it works correctly in Ubuntu Mate with the gtk tray icon because you can't remove appindicator support from Ubuntu because the system depends on it. Removing it uninstalls basically the whole DE...

JasonLG1979:
And ofc our symbolic icon still doesn't work with KDE...

pithos/plugins/notification_icon.py
```diff
@@ -201,6 +206,13 @@ def context_menu(self, widget, button, time, data=None):
                 data.show_all()
                 data.popup(None, None, None, None, 3, time)
 
+    def _on_icon_changed(self, prefs_dialog, icon):
+        self.settings['data'] = icon
+        if indicator_capable:
+            self.ind.set_icon_full(icon, '')
```
TingPing: Just `set_icon()`

pithos/plugins/notification_icon.py
```diff
@@ -92,20 +93,24 @@ def on_has_name_owner(bus, result):
...
+        if not self.settings['data']:
+            self.settings['data'] = 'io.github.Pithos-tray'
+
         logging.info('Checking if org.kde.StatusNotifierWatcher is owned')
         self.bus.call('org.freedesktop.DBus', '/', 'org.freedesktop.DBus',
                       'NameHasOwner', GLib.Variant('(s)', ('org.kde.StatusNotifierWatcher',)),
                       GLib.VariantType('(b)'), Gio.DBusCallFlags.NONE, -1, None, on_has_name_owner)
 
     def on_enable(self):
+        self.icon_changed_handler = self.preferences_dialog.connect('icon-changed', self._on_icon_changed)
```
TingPing: We can directly bind the GSetting rather than making a custom signal.

pithos/plugins/notification_icon.py
```diff
@@ -210,6 +222,92 @@ def on_disable(self):
...
+        self.set_resizable(False)
+        self.connect('response', self._on_response)
+
+        sub_title = Gtk.Label.new(_('Set the Notification Icon Type'))
+        sub_title.set_halign(Gtk.Align.CENTER)
+        self.icons_combo = Gtk.ComboBoxText.new()
+
+        icons = [
+            ('io.github.Pithos-tray', 'Full Color'),
+            ('io.github.Pithos-symbolic', 'Symbolic'),
```
TingPing: I think we should be using `io.github.Pithos-tray-symbolic`

pithos/plugins/notification_icon.py
```diff
@@ -210,6 +222,92 @@ def on_disable(self):
...
+            self.emit('icon-changed', active_id)
+            return
+
+        def on_dialog_response(dialog, response):
+            if response == Gtk.ResponseType.YES:
+                self.hide()
+                self.emit('icon-changed', active_id)
+            dialog.destroy()
+
+        message = (_(
```
TingPing: Do we need to have a warning really? Worst case it looks a little dumb but I think the user should understand it and nothing should break.

JasonLG1979: OK, I can get rid of it.

pithos/plugins/notification_icon.py
JasonLG1979: OK

pithos/plugins/notification_icon.py
JasonLG1979: set_icon is Deprecated

pithos/plugins/notification_icon.py
JasonLG1979: Why? That would literally just mean making a copy of the existing symbolic icon and naming it io.github.Pithos-tray-symbolic.

pithos/plugins/notification_icon.py
JasonLG1979: Well it's 6 of 1 thing 1/2 a dozen of another. If we bind the the combo to the setting we still have to connect a changed signal to the handler so the icon will change...

pithos/plugins/notification_icon.py
TingPing: Ah, maybe fill in a description then with 'Pithos'? Maybe it is used for accessibility or if icons are not supported?

JasonLG1979: The 2nd arg is the "icon description" whatever the hell that's suppose to mean? "Well it's sort of a vase looking pot with a music note on top of it."...

JasonLG1979: They both have a prop named icon_name so we can just do ```props.icon_name = icon``` for both and skip the icon description.

JasonLG1979:
Hey @TingPing I just noticed since things are async now when you enable a plugin with a pref the prefs button is not set sensitive. To access the prefs you have to click on a different plugin and click back on the other plugin. Then the prefs button is sensitive. Anyway what do you think about making the plugin base class a GObject so we can wire up a "prepared" signal. For that matter we can also wire "error" signal and possibly do other fancy stuff?

TingPing:
Making them a GObject sounds good.

TingPing:
At that point we could also make them implement https://lazka.github.io/pgi-docs/#Gio-2.0/classes/AsyncInitable.html

PR #470, opened 2017-05-17
First steps in making plugins GObjects

JasonLG1979:
This is the 1st step in making plugins GObjects. This allows us to potentially to do a bunch of neat stuff. For now this particular commit fixes a bug where when enabling a plugin with a prefs window the prefs button is not sensitive until you click off somewhere else and back on that plugin. This was caused because our plugins now load async.

### DIFF
--- a/pithos/PreferencesPithosDialog.py
+++ b/pithos/PreferencesPithosDialog.py
@@ -51,7 +51,8 @@ class PithosPluginRow(Gtk.ListBoxRow):
         self.switch.connect('notify::active', self.on_activated)
         self.switch.set_valign(Gtk.Align.CENTER)
         box.pack_end(self.switch, False, False, 2)
-        self.connect('grab-focus', self.on_grab_focus)
+        self.connect('grab-focus', self.set_prefs_btn)
+        self.plugin.connect('notify::enabled', self.on_enabled)
 
         if plugin.prepared and plugin.error:
             self.set_sensitive(False)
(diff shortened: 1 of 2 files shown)


pithos/PreferencesPithosDialog.py
```diff
@@ -51,15 +51,20 @@ def __init__(self, plugin):
         self.switch.connect('notify::active', self.on_activated)
         self.switch.set_valign(Gtk.Align.CENTER)
         box.pack_end(self.switch, False, False, 2)
-        self.connect('grab-focus', self.on_grab_focus)
+        self.connect('grab-focus', self.set_prefs_btn)
+        self.plugin.connect('enabled', self.on_enabled)
```
TingPing: __[flake8]__
* [W291] trailing whitespace

pithos/plugin.py
```diff
@@ -18,16 +18,23 @@
...
-    Gio
+    Gio,
+    GObject
 )
 
 
-class PithosPlugin:
+class PithosPlugin(GObject.Object):
+    __gtype_name__ = 'PithosPlugin'
+    __gsignals__ = {
```
TingPing: Why make it a signal rather than a property, you still get change notifications but you get the benefit of it being a toggle and can be bound.

JasonLG1979: I can do that. I really don't want to bind it to anything ATM though. Prob just connect to the notify signal.

JasonLG1979:
Better? 'enabled' is now a prop not a signal.

JasonLG1979:
@TingPing how about merging this?

JasonLG1979:
Bump. I did what you asked. How about merging this @TingPing?

PR #473, opened 2017-05-26
Update to copyright year

RyanNerd:
Small change in about dialog that updates the copyright year.

### DIFF
--- a/data/ui/AboutPithosDialog.ui
+++ b/data/ui/AboutPithosDialog.ui
@@ -4,7 +4,7 @@
   <requires lib="gtk+" version="3.14"/>
   <template class="AboutPithosDialog" parent="GtkAboutDialog">
     <property name="program_name">Pithos</property>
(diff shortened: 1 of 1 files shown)


TingPing:
Only the start date of a copyright claim matters, beyond that it lasts the year of the author + whatever in the US so there is no point in mentioning it.

JasonLG1979:
Every time Mickey Mouse comes up to go into the public domain they extend the copyright term(no joke, true story). It's like life of the author + 75 years now I think.

RyanNerd:
So should the year even be mentioned then? Currently, it is set in the app as 2012. Here's some more [Disney copyright insanity](https://www.techdirt.com/articles/20170525/10481837456/piracy-killing-hollywood-so-bad-that-disney-made-more-money-2016-than-any-studio-ever.shtml) that is also true.

JasonLG1979:
> So should the year even be mentioned then? Currently, it is set in the app as 2012.

I think you may still have to declare a starting year? If anything it's convention.

PR #477, opened 2017-05-31
connect to is-active mediakeys plugin

JasonLG1979:
Mate doesn't send focus events like GNOME which results in the mediakeys never actually binding this should fix that by connecting to the 'notify::is-active' prop change signal instead which works as expected on both GNOME and Mate.

### DIFF
--- a/pithos/plugins/mediakeys.py
+++ b/pithos/plugins/mediakeys.py
@@ -53,9 +53,8 @@ class MediaKeyPlugin(PithosPlugin):
             None,
         )
 
-    def update_focus_time(self, widget, event, userdata=None):
-        if event.changed_mask & Gdk.WindowState.FOCUSED and \
-           event.new_window_state & Gdk.WindowState.FOCUSED:
+    def update_active(self, *ignore):
+        if self.window.is_active():
             self.grab_media_keys()
 
     def mediakey_signal(self, proxy, sender, signal, param, userdata=None):
@@ -130,7 +129,7 @@ class MediaKeyPlugin(PithosPlugin):
 
     def on_enable(self):
         if self.method == 'dbus':
-            self.focus_hook = self.window.connect('window-state-event', self.update_focus_time)
+            self.active_hook = self.window.connect('notify::is-active', self.update_active)
             self.mediakey_hook = self.mediakeys.connect('g-signal', self.mediakey_signal)
             logging.info('Bound media keys with DBUS {}'.format(self.mediakeys.props.g_interface_name))
         elif self.method == 'keybinder':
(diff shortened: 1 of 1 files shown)


JasonLG1979:
@TingPing have you had a chance to test this for yourself?

TingPing:
This is functional but it seems to run a lot more frequently than necessary.

JasonLG1979:
> This is functional but it seems to run a lot more frequently than necessary.

I really don't see the harm in that, grabing the keys is a async function? It's the only way I can figure out to make it work properly on all DE's that support DBus mediakeys.

TingPing:
Yea, I sort of want to understand what MATE is doing before working around it personally.

JasonLG1979:
> Yea, I sort of want to understand what MATE is doing before working around it personally.

Basically what we have never passes ever. I don't know if that's by design on their part or a bug?

JasonLG1979:
Totally off topic but @TingPing the original author of the media player extension turned over ownership to the repo to me so if you ever wanted me to clean up the code on that I wouldn't say no,lol!!!

TingPing:
@JasonLG1979 Congrats, may you be a better maintainer than me :P

Next `gjs` release though I really want to write some JS, I even thought about rewriting Pithos for fun :P

JasonLG1979:
> Next gjs release though I really want to write some JS, I even thought about rewriting Pithos for fun :P

Extensions are the redheaded stepchild of the Shell. It's really stupid there's really very little to no docs, no API and the actual preferred method to mess with the shell is monkey patching... It's that bad,lol!!!

JasonLG1979:
> Congrats, may you be a better maintainer than me :P

Thanks!!! There's really not that much to maintain. I've been the only contributor for months. Basically it's just me tweaking stuff trying to make it better, responding to bug reports and feature requests. One of these days i'm going to add Pithos ratings support. i'm still trying to figure out how to do it so it doesn't look stupid.

TingPing:
> Extensions are the redheaded stepchild of the Shell. It's really stupid there's really very little to no docs, no API and the actual preferred method to mess with the shell is monkey patching... It's that bad,lol!!!

Yea extensions are not great, I've been working with the gjs maintainer and we have *some* docs online now: https://devdocs.baznga.org/ and there are plans to further improve it.

Anyway the language itself has a lot of nice features coming, ignoring the gnome-shell bindings.

JasonLG1979:
> Yea extensions are not great, I've been working with the gjs maintainer and we have some docs online now: https://devdocs.baznga.org/ and there are plans to further improve it.
> 
> Anyway the language itself has a lot of nice features coming, ignoring the gnome-shell bindings.

Nice!!! Up until this point I've been using the python docs as a reference and then by trial and error translating it to Gjs.

JasonLG1979:
@TingPing What about this? I'll test against Ubuntu Mate 17.10 beta 1 and see if anything has changed.

TingPing:
I still think it is a bad workaround for a broken window manager.

JasonLG1979:
Media keys still don't work as they should... I'll check to see if there are any outstanding bugs relating to this issue.

PR #481, opened 2017-07-03
Add an option to hide on close in the mpris plugin

JasonLG1979:
This makes no attempt to find MPRIS clients in any way. It simply provides a switch for the use to decide. When the MPRIS plugin is disabled the default window behavior is restored.

Generally it's pretty safe to hide on close even if there is no MPRIS client. Worst case you have to hit the launcher to get the window back.

The alternative is trying to sniff out 75 different MPRIS clients various ways.

### DIFF
--- a/pithos/plugins/mpris.py
+++ b/pithos/plugins/mpris.py
@@ -23,7 +23,8 @@ import math
 
 from gi.repository import (
     GLib,
-    Gio
+    Gio,
+    Gtk
 )
 from .dbus_util.DBusServiceObject import (
     DBusServiceObject,
@@ -50,6 +51,7 @@ class MprisPlugin(PithosPlugin):
                 logging.warning('Failed to create DBus mpris service: {}'.format(e))
                 self.prepare_complete(error='Failed to create DBus mpris service')
             else:
+                self.preferences_dialog = MprisPluginPrefsDialog(self.window, self.settings)
                 self.prepare_complete()
 
     def on_enable(self):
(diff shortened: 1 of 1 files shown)


pithos/plugins/mpris.py
```diff
@@ -830,3 +832,60 @@ def PropertiesChanged(self, interface, changed, invalidated):
...
+        self.settings = settings
+        self.window_delete_handler = None
+
+        box = Gtk.Box()
+        label = Gtk.Label()
+        label.set_markup('<b>{}</b>\n{}'.format(_('Hide Pithos on Close'), _('Instead of Quitting')))
+        label.set_halign(Gtk.Align.START)
+        box.pack_start(label, True, True, 4)
+
+        self.switch = Gtk.Switch()
```
TingPing: __[flake8]__
* [W291] trailing whitespace

pithos/plugins/mpris.py
```diff
@@ -830,3 +832,60 @@ def PropertiesChanged(self, interface, changed, invalidated):
...
+        content_area.show_all()
+
+    def on_close(self, window, event):
+        window.hide()
+        return True
+
+    def on_activated(self, *ignore):
+        if self.switch.get_active():
+            self.settings['data'] = 'True'
+            self.delete_callback_handle = self.pithos.connect('delete-event', self.on_close)
```
TingPing: __[flake8]__
* [W291] trailing whitespace

JasonLG1979:
@TingPing What about this?

JasonLG1979:
With what's soon to going on with the notification icon plugin this might be a nice feature.

TingPing:
On a side note in gnome it seems like mpris show loses the window position.

JasonLG1979:
> On a side note in gnome it seems like mpris show loses the window position.

That's because activating an app over DBus calls present. If you want it to not loose window position you should override present and set the window position there.

JasonLG1979:
```python
    def bring_to_top(self, *ignore):
        self.present()

    def present(self):
        self.set_initial_pos()
        Gtk.Window.present(self)
```

Should do the trick.

JasonLG1979:
Or even better:
```python

    def bring_to_top(self, *ignore):
        timestamp = Gtk.get_current_event_time()
        self.present_with_time(timestamp)

    def present_with_time(self, timestamp):
        self.set_initial_pos()
        Gtk.Window.present_with_time(self, timestamp)

    def present(self):
        self.set_initial_pos()
        Gtk.Window.present(self)
```

JasonLG1979:
That would make it so that no matter if a MPRIS client call activate or the actual raise method they would be calling the exact same function pretty much.

PR #488, opened 2017-07-26
Tweak for notifications in GNOME Shell >= 3.24

JasonLG1979:
Icons/images in GNOME Shell 3.24 notifications are really small (16px?). It looks better to just send the Pithos symbolic icon than a cover and that seems to be what other players are doing (Lollypop, Rhythmbox). This aslo changes the notification text to match the stock MPRIS widget.

### DIFF
--- a/pithos/plugins/notify.py
+++ b/pithos/plugins/notify.py
@@ -29,6 +29,7 @@ class NotifyPlugin(PithosPlugin):
     notification = None
     supports_actions = False
     escape_markup = False
+    no_cover = False
 
     def on_prepare(self):
         def on_notify_init_finish(notification, server_info, caps, error=None):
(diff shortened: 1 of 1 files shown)


pithos/plugins/notify.py
```diff
@@ -51,8 +52,11 @@ def on_notify_init_finish(notification, server_info, caps, error=None):
                 version = server_info['version'].split('.')
                 if server_info['name'] == 'gnome-shell' and len(version) >= 2:
                     major_version, minor_version = (int(x) if x.isdigit() else 0 for x in version[0:2])
-                    if major_version == 3 and minor_version >= 20:
-                        has_built_in_mpris = True
+                    if major_version == 3:
+                        if minor_version >= 20:
+                            has_built_in_mpris = True
+                        if minor_version >= 24:
+                            self.no_cover = True
```
TingPing: __[flake8]__
* [W291] trailing whitespace

TingPing:
I do remember really small art icons but as of right now on F26's version of GNOME 3.24 the album art is larger and clearly visible.

JasonLG1979:
Hey, @TingPing while you're in a merging mood how about you check out my last couple PR's?

TingPing:
I am not a fan of this idea. I'd honestly rather start a conversation with gnome-shell folks on why they made this silly change.

JasonLG1979:
I'm OK with that. My thought is that they're just not a fan of images in notifications in general and not a fan of media player notifications at all. If GNOME Music is any indication they see media player notifications as redundant with the built in MPRIS support.

JasonLG1979:
I see it as just a matter of time before the only thing you can send in a GNOME notification is a icon name.

JasonLG1979:
@TingPing, so no go on this? I'm trying to figure out what I should be working on for 1.4?

TingPing:
I do not think this change is needed.

JasonLG1979:
Right on. I'm in the process of downloading Ubuntu Mate, but it's taking a while because of my crappy internet...

PR #492, opened 2017-08-12
Add a plugin to inhibit idle(screensaver)/auto suspend

JasonLG1979:


TingPing:
Should it really be a plugin?

JasonLG1979:
> Should it really be a plugin?

If by that you mean "Is this a useful feature?" I'd say yes. If by that you mean "Why make a built in Gtk function a plugin?" I'd say because that's pretty much where we put most of our user options.

pithos/plugins/inhibit_screensaver.py
```diff
@@ -0,0 +1,79 @@
...
+        idle = Gtk.ApplicationInhibitFlags.IDLE
+
+        self._cookie = self._pithos_app.inhibit(
+            self.window,
+            suspend | idle,
+            'Inhibit Screensaver plugin enabled',
+        )
+
+    def _uninhibit(self):
+        if self._cookie and self._cookie != 0:
```
TingPing: redundant

TingPing:
Well I was just thinking it has no dependencies and seems logical as default behavior but there is always somebody who would probably want to disable it.

pithos/plugins/inhibit_screensaver.py
JasonLG1979: fixed.

JasonLG1979:
> Well I was just thinking it has no dependencies and seems logical as default behavior but there is always somebody who would probably want to disable it.

I'm not sure I'd make it default behavior. I like it better as an option. I kinda see it as a complimenting, almost opposite option to the Screensaver Pause plugin. Which I see as a better default behavior.

PR #493, opened 2017-08-13
Override present and present_with_time

JasonLG1979:
This should make sure our position is never lost.

### DIFF
--- a/pithos/plugins/notification_icon.py
+++ b/pithos/plugins/notification_icon.py
@@ -194,11 +194,10 @@ class PithosNotificationIcon(PithosPlugin):
             self.statusicon.set_tooltip_text("{} by {}".format(song.title, song.artist))
 
     def _toggle_visible(self, *args):
-        self.window.set_visible(not self.window.get_visible())
-
-        if self.window.get_visible(): # Ensure it's on top
+        if self.window.get_visible():
+            self.window.hide()
+        else:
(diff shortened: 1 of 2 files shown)


TingPing:
Can probably remove our other call to `self.set_initial_pos()` on Window creation then?

TingPing:
Also maybe rename that function to `restore_position()` or something now.

JasonLG1979:
> Can probably remove our other call to self.set_initial_pos() on Window creation then?



> Also maybe rename that function to restore_position() or something now.

Ok, sounds good. I'll amend the PR.

JasonLG1979:
Done.

pithos/plugins/notification_icon.py
```diff
@@ -194,11 +194,10 @@ def song_changed(self, window, song):
             self.statusicon.set_tooltip_text("{} by {}".format(song.title, song.artist))
 
     def _toggle_visible(self, *args):
-        self.window.set_visible(not self.window.get_visible())
-
-        if self.window.get_visible(): # Ensure it's on top
+        if self.window.get_visible():
+            self.window.hide_on_delete()
```
TingPing: Just use `self.window.hide()`. That method is only a convenience thing for directly connecting signals.

JasonLG1979: Done.

PR #494, opened 2017-08-13
Check to see if we're running in GNOME Shell and if so what version in Notification Icon Plugin

JasonLG1979:
Tray icons are not supported in GNOME Shell 3.26 +. This disable the notification icon plugin if we are running in GNOME Shell version 3.26+ and don't have appindicator support and tell the user why it was disabled.

### DIFF
--- a/pithos/plugins/notification_icon.py
+++ b/pithos/plugins/notification_icon.py
@@ -60,6 +60,26 @@ class PithosNotificationIcon(PithosPlugin):
             self.prepare_complete(error='DBus failed and is required for notification icon')
             return
 
+        def on_gnome_shell_proxy_finished(source, result, data):
+            try:
+                gnome_shell_proxy = Gio.DBusProxy.new_finish(result)
+            except GLib.Error as e:
+                logging.exception(e)
+                self.prepare_complete(error='Failed to find DBus service')
+            else:
+                if (gnome_shell_proxy.get_name_owner() and
+                        'ShellVersion' in gnome_shell_proxy.get_cached_property_names()):
+                    version = gnome_shell_proxy.get_cached_property('ShellVersion').unpack()
+                    logging.info('GNOME Shell version: {}'.format(version))
+                    version = version.split('.')
+                    major_version, minor_version = (int(x) if x.isdigit() else 0 for x in version[0:2])
+                    if minor_version > 24 or major_version > 3:
+                        self.prepare_complete(error='Tray Icons are no longer supported in GNOME Shell version 3.26 +')
+                    else:
+                        self.prepare_complete()
+                else:
+                    self.prepare_complete()
+
         def on_has_name_owner(bus, result):
             try:
                 is_owned = bus.call_finish(result)[0]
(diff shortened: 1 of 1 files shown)


pithos/plugins/notification_icon.py
```diff
@@ -91,7 +111,18 @@ def on_has_name_owner(bus, result):
                 self.prepare_complete()
             else:
                 # No fancy tray here
-                self.prepare_complete()
+                logging.info('Checking if org.gnome.Shell is owned and version')
+                Gio.DBusProxy.new(
+                    self.bus,
+                    Gio.DBusProxyFlags.NONE,
+                    None,
+                    'org.gnome.Shell',
```
TingPing: I think we discussed this before, but was there really no better way to find this information? I really don't want to open such as huge hole in the Flatpak sandbox to talk to `org.gnome.Shell`.

JasonLG1979: No, not that I can tell beyond sniffing the DE in python itself which is not reliable and I'm not sure how that would even work in Flatpack?

JasonLG1979: The other time this came up was with the notify plugin. In that case the notification interface itself returns a version.

JasonLG1979: The only real alternatives are either just skip the check and always assume one way or the other or just completely disable the notification icon plugin altogether.

TingPing: Hmm how about this:

```python
if os.environ['XDG_CURRENT_DESKTOP'] == 'GNOME':
     # Check org.gnome.Shell for version, if fails assume not supported
else:
     # Previous behavior
```

JasonLG1979: Last time I checked several DE's reported GNOME as the DE in

```python
os.environ['XDG_CURRENT_DESKTOP']
```
As I said sniffing DE's in python isn't reliable.

TingPing: oh =(

JasonLG1979: Just google 'XDG_CURRENT_DESKTOP' and you'll see a billion bug reports of apps using that to sniff the DE and failing.

JasonLG1979: Can you not just open up flatpak just enough to read that prop and that's all?

JasonLG1979: I'm surprised that GNOME Shell doesn't expose that to Flatpak by default. You'd think that would be nice thing to know?

TingPing: Currently the Flatpak dbus support is based upon bus names only. I would love to lock it down by property or interface but that requires work there.

JasonLG1979: Looks like you got some work to do then ;P

In my mind you'd want to be able to lock it down to the point that you could even specify if the app is even allowed to set read/write props.

JasonLG1979:
Well no need for this anymore ;)

PR #495, opened 2017-08-19
We don't have to lie about being able to seek any longer

JasonLG1979:
The visibility of the position slider in https://github.com/JasonLG1979/gnome-shell-extensions-mediaplayer was decoupled from the CanSeek prop a while ago. As long as a Player provides a length in the metadata and can provide a valid postion the slider is visible. CanSeek determines if the slider is reactive. For Pithos that means the slider is visible but can't be moved by the user.


JasonLG1979:
And this to please :)

PR #500, opened 2017-08-26
Use trackGain from Pandora to fake a ReplayGain Plugin

JasonLG1979:
Use the ReplayGain Gstreamer Elements in conjunction with the trackGain from Pandora to create a "Auto Volume Normalization" plugin.


pithos/plugins/auto_volume_normalization.py
```diff
@@ -0,0 +1,60 @@
...
+    description = _('Normalize apparent volume')
+    _song_change_handler = None
+
+    def on_prepare(self):
+        self.prepare_complete()
+
+    def on_enable(self):
+        self._song_change_handler = self.window.connect('song-changed', self._on_song_changed)
+        self.window.rglimiter.set_property('enabled', True)
+        if self.window.current_song is not None:
```
TingPing: __[flake8]__
* [W291] trailing whitespace

TingPing:
Huh, interesting they have that property. Do you know if any of their players use this?

JasonLG1979:
No idea? The official players are black boxes. Although testing with with various genres from different times in the "loudness wars" seems to confirm that the prop's intended purpose is gain normalization.

TingPing:
Is there a downside to always doing this?

JasonLG1979:
ReplayGain works by 1st running the audio through a soft knee limiter then dropping the over all gain to provide headroom for adjustments. Then it applies the adjustment. It's a trade off. "Louder" sounding songs will be quieter but quieter songs will sound louder. It's most noticeable when listening to stations with songs that span different decades, as pretty much the newer the master the louder it sounds. The drawback is that the peak volume or apparent volume of the songs on average will drop a little bit. It's not something I'd do by default as some people may not want that.

TingPing:
>  It's not something I'd do by default as some people may not want that.

From my perspective normalized volume is what more people would want, nobody likes constantly adjusting volume. An option to opt out for some purist would be reasonable though.

JasonLG1979:
My concern is that if it is enabled by default that we'll be flooded with bug reports about the sudden volume drop.

JasonLG1979:
> nobody likes constantly adjusting volume.

On stations with more "homogeneous" songs you shouldn't need to adjust the volume as they'd probably have about the same apparent volume. Like I said the apparent volume difference between songs is going to be the most pronounced with stations that mix genes and eras. For example I have a Classic Rock station that has songs from anywhere from the late 50's to the late 80's. For that station volume normalization is a plus. I also have a Modern Rock station that only plays stuff from the past 5-10 years to the present. For that station volume normalization is pointless because all the songs have about the same apparent volume anyway. For that station I'd like to actually get full digital scale.

JasonLG1979:
There's also the fact that no matter how good the processing is, audio processing is going to add at least a little bit of distortion. The objectively "best" sound quality is going to be the audio with the least processing. I know that we're not going to get audiophile sound from a Pandora stream but to me the default should unprocessed audio(by use at least).

pithos/plugins/auto_volume_normalization.py
```diff
@@ -0,0 +1,60 @@
...
+            self.window.rgvolume.set_property('fallback-gain', 0.0)
+            self.window.rglimiter.set_property('enabled', False)
+            dialog.destroy()
+
+        dialog = Gtk.MessageDialog(
+            parent=self.window.prefs_dlg,
+            flags=Gtk.DialogFlags.MODAL,
+            type=Gtk.MessageType.WARNING,
+            buttons=Gtk.ButtonsType.OK,
+            text=_('Please Lower The Volume'),
```
TingPing: This dialog is modal, so you can't really lower the volume in Pithos =\

JasonLG1979: You can lower your system volume

JasonLG1979:
I just amended the PR to help address some of the objective sound quality issues of processing the stream by making sure all audio conversion is done at the highest quality.

pithos/plugins/auto_volume_normalization.py
JasonLG1979: What about pausing playback to give the use the chance to lower the volume?

JasonLG1979: There now it Pauses Pithos and tells the user to lower the volume before resuming playback.

TingPing:
@JasonLG1979 Mind making a patch to sort the plugin list alphabetically or something? Getting a bit long.

JasonLG1979:
sure.

JasonLG1979:
Done

PR #501, opened 2017-08-29
Sort Plugins Alphabetically

JasonLG1979:


pithos/plugins/auto_volume_normalization.py
```diff
@@ -58,7 +58,7 @@ def _volume_warning_dialog(self):
             buttons=Gtk.ButtonsType.OK,
             text=text,
             secondary_text=_(
-                'Please lower the volume before resumming playback,\n'
+                'Please lower the volume before resuming playback,\n'
```
TingPing: Make it another commit.

JasonLG1979: Done

PR #503, opened 2017-08-31
song.title should not be a property

JasonLG1979:
If song.title is a decorated property with a getter that means that the 1st thing to access it has to potentially do a blocking network call. It just so happens that the 1st thing to access song.title does so from the main thread. Doing blocking calls from the main thread is ofc bad. This instead moves checking the title to init to make sure that any blocking calls are done from a worker thread.

I never really noticed the UI blocking until recently because I've had issues with my internet connection. So I threw a ```print(threading.get_ident())``` in the title prop and sure enough it was doing blocking I/O in the main thread.

### DIFF
--- a/pithos/pandora/pandora.py
+++ b/pithos/pandora/pandora.py
@@ -433,7 +433,17 @@ class Song:
     def __init__(self, pandora, d, playlist_time):
         self.pandora = pandora
         self.playlist_time = playlist_time
-
+        self.is_ad = None  # None = we haven't checked, otherwise True/False
+        self.tired = False
+        self.message = ''
+        self.duration = None
+        self.position = None
+        self.bitrate = None
+        self.start_time = None
+        self.finished = False
+        self.feedbackId = None
+        self.bitrate = None
+        self.artUrl = None
         self.album = d['albumName']
         self.artist = d['artistName']
         self.trackToken = d['trackToken']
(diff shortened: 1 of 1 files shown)


TingPing:
But now something will get the wrong empty title?

JasonLG1979:
No. title will always either be the value from the explorer url or songName. Nothing will get an empty string because that function is called during class creation in a worker thread before anything else tries to access title.

JasonLG1979:
title starts as an empty string because it's just bad form to create class vars outside of the init IMHO. In my mind all "self"'s should be declared in init.

TingPing:
Alright, should improve the commit message though, title is clearly still a property it just no longer does IO in the getter.

JasonLG1979:
Look at it again, it's not a property it's just a normal instance variable. There is no "getter".

JasonLG1979:
If you like it doesn;t really need to be inside a function it can be naked in init also ```__repr__``` should probably return self.title instead of self.songName like so: 

```python
    def __repr__(self):
        return '<{}.{} {} "{}" by "{}" from "{}">'.format(
            __name__,
            __class__.__name__,
            self.trackToken,
            self.title,
            self.artist,
            self.album,
        )
```

We don't use songName anywhere else in Pithos.

TingPing:
>  it's not a property it's just a normal instance variable. There is no "getter".

At least by my definition an instance variable is a property by default, it just has no custom getter or setter.

Your call on `__repr__`.

JasonLG1979:
What about skipping the function and just putting it in init?

TingPing:
Its ugly eitherway but there is no real benefit to it being a function atm.

JasonLG1979:
There it's all out in the open and I clarified the commit message a bit. I also moved all the instance class vars up above all the ugliness.

PR #504, opened 2017-09-01
Ensure that the notification icon and journald logging plugin prefs windows are not destroyed when they recieve a delete event

JasonLG1979:


pithos/plugins/journald_logging.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, parent, settings):
         self.pithos_window = parent
         self.settings = settings
         self.set_resizable(False)
-        self.connect('response', self._on_response)
+        self.connect('delete-event', lambda *ignore: self.do_response(Gtk.ResponseType.CANCEL))
```
TingPing: Probably just want `self.response()`

JasonLG1979: OK

JasonLG1979: Well i guess I din't understand what you mean?
```self.connect('delete-event', lambda *ignore: self.do_response(self.response())```

TingPing:
```python
self.connect('delete-event', lambda *ignore: self.response(Gtk.ResponseType.CANCEL))
 ```

JasonLG1979: right on

JasonLG1979: Nope that didn't work. I got:

```gtk_container_foreach: assertion 'GTK_IS_CONTAINER (container)' failed``

When I tried to reopen the window and a segfault and crash when I clicked on it.

JasonLG1979: @TingPing the problem I'd guess is that ```self.response``` doesn't return True.

JasonLG1979: ```self.connect('delete-event', lambda *ignore: self.response(Gtk.ResponseType.CANCEL) or True)``` seems to work though.

JasonLG1979:
There what do you think of that?

JasonLG1979:
There you go @TingPing all rebased against master ready to merge :)

PR #506, opened 2017-09-03
Keep pandora.stations list in sync

JasonLG1979:
We create and delete stations but don't append or remove them from pandora.stations as we should. this fixes that.


JasonLG1979:
There you go @TingPing all rebased against master ready to merge :)

PR #509, opened 2017-09-04
We can't use the GstAudio constants

JasonLG1979:
As much as I would prefer to use the GstAudio.AUDIO_RESAMPLER_QUALITY_MAX and GstAudio.AudioResamplerFilterMode.FULL constants, they don't exist in the version of Gstreamer that ships with *buntu 16.04. Trying to use them in *buntu 16.04 crashes Pithos.

Instead just set "quality" and "sinc-filter-mode" to the values the constants would be if they existed...

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -52,6 +52,14 @@ try:
 except ImportError:
     pacparser = None
 
+# Older versions of Gstreamer may not have these constants
+try:
+    RESAMPLER_QUALITY_MAX = GstAudio.AUDIO_RESAMPLER_QUALITY_MAX
+    RESAMPLER_FILTER_MODE_FULL = GstAudio.AudioResamplerFilterMode.FULL
+except AttributeError:
+    RESAMPLER_QUALITY_MAX = 10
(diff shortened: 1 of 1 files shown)


pithos/pithos.py
```diff
@@ -300,8 +299,8 @@ def init_core(self):
         self.equalizer = Gst.ElementFactory.make("equalizer-10bands", "equalizer-10bands")
         audioconvert = Gst.ElementFactory.make("audioconvert", "audioconvert")
         audioresample = Gst.ElementFactory.make("audioresample", "audioresample")
-        audioresample.set_property("quality", GstAudio.AUDIO_RESAMPLER_QUALITY_MAX)
-        audioresample.set_property("sinc-filter-mode", GstAudio.AudioResamplerFilterMode.FULL)
+        audioresample.set_property("quality", 10)
```
TingPing: At least define them as usable names elsewhere, and maybe leave a comment saying these constants are missing in X version of the gstreamer bindings.

JasonLG1979: OK

JasonLG1979:
Something like that @TingPing ?

PR #522, opened 2017-10-09
Raise the priority of the GLib.idle_add's in the worker to GLib.PRIORITY_DEFAULT

JasonLG1979:
As it sits now the entry priority of worker threads is really low. Worker threads have to wait for time to refresh in the UI for example because it's a higher priority.

This commit has the effect of making things that use the worker thread generally more "snappy". Songs load faster in the UI, station searches are faster, song ratings are faster.


TingPing:
:-1: Logic should always be lower than the UI as we never want to block rendering.

PR #541, opened 2017-11-20
Work Around Broken Notification Servers

JasonLG1979:
Deepin's notification system is broken, see: https://github.com/gnumdk/lollypop/issues/1203
This will stop exceptions by ignoring parameters and send a warning message about the broken signal.

This also fixes a workaround for GNOME sending multiple redundant 'NotificationClosed' signals.

The original workaround breaks actions in KNotifications because KNotifications does not increment the notification id's as they should. Instead each app has it's own id. Including signal_value in self._last_signal make it work for both.

### DIFF
--- a/pithos/plugins/dbus_util/GioNotify.py
+++ b/pithos/plugins/dbus_util/GioNotify.py
@@ -18,6 +18,8 @@
 # <https://github.com/JasonLG1979/possibly-useful-scraps/wiki/GioNotify>
 # for documentation.
 
+import logging
+
 from enum import Enum
 
 from gi.repository import GLib, GObject, Gio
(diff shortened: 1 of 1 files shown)


pithos/plugins/dbus_util/GioNotify.py
```diff
@@ -158,20 +163,58 @@ def set_hint(self, key, value):
...
-        # In GNOME Shell at least this stops multiple
-        # redundant 'NotificationClosed' signals from being emmitted.
-        if (id, signal_name) == self._last_signal:
-            return
-        self._last_signal = id, signal_name
-        if signal_name == 'ActionInvoked':
-            self.emit('action-invoked', signal_value)
-            self._callbacks[signal_value]()
+        try:
+            id, signal_value = parameters.unpack()
```
TingPing: I know this was used before but can you avoid overriding `id` which is a builtin function.

JasonLG1979: No problem

PR #548, opened 2017-11-24
Remember the windows size and if it's maximized

JasonLG1979:
This also debounces the window configure events as to not hammer the settings and takes into acount that getting the windows position in Wayland is not possible.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -361,6 +361,9 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.buffering_timer_id = 0
         self.ui_loop_timer_id = 0
         self.playlist_update_timer_id = 0
+        self.configure_event_timer_id = 0
+        display = self.props.screen.get_display()
+        self.running_in_x = type(display).__name__.endswith('X11Display')
         self.worker = GObjectWorker()
 
         try:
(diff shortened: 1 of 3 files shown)


pithos/pithos.py
```diff
@@ -1460,22 +1463,63 @@ def restore_position(self):
         if not x is None and not y is None:
             self.move(int(x), int(y))
 
+    def restore_size(self):
+        """ resizes window to size stored in preferences """
+        w, h = self.settings['win-size']
+        if not w is None and not h is None:
```
TingPing: __[Flake8]__

test for object identity should be 'is not'

pithos/pithos.py
```diff
@@ -1460,22 +1463,63 @@ def restore_position(self):
...
+            # does not work in Wayland.
+            if self.running_in_x:
+                x, y = widget.get_position()
+                self.settings.set_value('win-pos', GLib.Variant('(ii)', (x, y)))
+            w, h = widget.get_size()
+            self.settings.set_value('win-size', GLib.Variant('(ii)', (w, h)))
+            return False
+        # Debouncer. Don't hammer settings.
+        # Store the settings after the configure
+        # events stop comming.
```
TingPing: __[Flake8]__

trailing whitespace

pithos/pithos.py
```diff
@@ -1460,22 +1463,63 @@ def restore_position(self):
         if not x is None and not y is None:
             self.move(int(x), int(y))
 
+    def restore_size(self):
+        """ resizes window to size stored in preferences """
+        w, h = self.settings['win-size']
+        if w is not None and h is not None:
```
TingPing: These can never be `None`.

pithos/pithos.py
```diff
@@ -1460,22 +1463,63 @@ def restore_position(self):
...
+
+    def restore_maximized(self):
+        if self.settings['win-maxed']:
+            self.maximize()
+
+    def restore_window(self):
+        # Getting the window position
+        # does not work in Wayland.
+        if self.running_in_x:
+            self.restore_position()
```
TingPing: I believe it is just ignored on Wayland so there is no harm in always calling it?

JasonLG1979: It's not ignored, It returns a value, just not the correct value, if I remember for me it was always returning 15, 23 no matter what.

pithos/pithos.py
JasonLG1979: I was just following what you did with the position.

TingPing: Past me was an idiot =)

pithos/pithos.py
TingPing: Strange but alright.

pithos/pithos.py
```diff
@@ -1460,22 +1463,63 @@ def restore_position(self):
...
+            if self.running_in_x:
+                x, y = widget.get_position()
+                self.settings.set_value('win-pos', GLib.Variant('(ii)', (x, y)))
+            w, h = widget.get_size()
+            self.settings.set_value('win-size', GLib.Variant('(ii)', (w, h)))
+            return False
+        # Debouncer. Don't hammer settings.
+        # Store the settings after the configure
+        # events stop comming.
+        if self.configure_event_timer_id:
```
TingPing: So I honestly question if this actually improves anything. Making new GSources is a relatively expensive operation as I understand it so you replaced writing to DConf repeatedly which is a queued write operation and is fast with a slower global glib lock to a queued operation.

Now I might be wrong but without profiling I don't think we should make our code much uglier unless its provably a performance win.

JasonLG1979: Another reason for this is that it gives the window time to actually finish resizing before we check the size. If you go straight from a configure event to checking the size you can end up with a stale size.

JasonLG1979:
I'm not married to any of this. It for sure doesn't need to hold up the bug fix release.

JasonLG1979:
It's actually kinda a PITA to get consistent behavior between X and Wayland and that's just on GNOME. Who know how other DE's will handle windows in either case?

pithos/pithos.py
JasonLG1979: You still get configure events so it's set to the invalid value. The end result is that if you switch from an X session to a Wayland session and back again to an X session it basically wipes your setting. Which isn't a huge deal. Remembering this stuff is a waste of time anyway... But generally we could save a tiny bit of resources in Wayland by just disconnecting the handler altogether. At the very least we should block the handler when we restore the position.

PR #578, opened 2018-11-23
Further simplifiy/improve the notifiy plugin

JasonLG1979:
Remove unused logging import
Don't keep track of notification id's
Add a fallback icon

Works fine on GNOME. Getting ready to test this on XFCE, KDE, MATE, Cinnamon and Pantheon and will makes further tweaks if needed.

### DIFF
--- a/pithos/application.py
+++ b/pithos/application.py
@@ -88,12 +88,9 @@ class PithosApplication(Gtk.Application):
         action.connect("activate", lambda action, param: self.window.next_song())
         self.add_action(action)
 
-        action = Gio.SimpleAction.new("play", None)
-        action.connect("activate", lambda action, param: self.window.user_play())
-        self.add_action(action)
-
-        action = Gio.SimpleAction.new("pause", None)
-        action.connect("activate", lambda action, param: self.window.user_pause())
+        # Needed for notifications to function as expected in KDE.
+        action = Gio.SimpleAction.new("activate", None)
+        action.connect("activate", lambda action, param: self.activate())
         self.add_action(action)
 
         if Gtk.get_major_version() > 3 or Gtk.get_minor_version() >= 20:
--- a/pithos/plugins/notify.py
+++ b/pithos/plugins/notify.py
@@ -13,7 +13,6 @@
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
-import logging
 
 from gi.repository import Gio
 
@@ -25,7 +24,8 @@ class NotifyPlugin(PithosPlugin):
     description = 'Shows notifications on song change'
 
     _app = None
-    _last_id = ''
+    _app_id = None
+    _fallback_icon = None
 
     def on_prepare(self):
         # We prefer the behavior of the fdo backend to the gtk backend
@@ -35,40 +35,44 @@ class NotifyPlugin(PithosPlugin):
             os.environ['GNOTIFICATION_BACKEND'] = 'freedesktop'
 
         self._app = Gio.Application.get_default()
+        self._app_id = self._app.get_application_id()
+        self._fallback_icon = Gio.ThemedIcon.new('audio-x-generic')
         self.prepare_complete()
 
     def on_enable(self):
-        self.song_change_handler = self.window.connect('song-changed', self.send_notification)
-        self.state_change_handler = self.window.connect('user-changed-play-state', self.send_notification)
+        self._song_change_handler = self.window.connect('song-changed', self.send_notification)
+        self._shutdown_handler = self._app.connect('shutdown', lambda app: app.withdraw_notification(self._app_id))
 
     def send_notification(self, window, *ignore):
         if window.is_active():
-            return
-
-        song = window.current_song
-        # This matches GNOME-Shell's format
-        notification = Gio.Notification.new(song.artist)
-        notification.set_body(song.title)
-        if song.artUrl:
-            notification.set_icon(Gio.FileIcon.new(Gio.File.new_for_uri(song.artUrl)))
-
-        if window.playing:
-            notification.add_button(_('Pause'), 'app.pause')
+            # GNOME-Shell will auto dismiss notifications
+            # when the window becomes "active" but other DE's may not (KDE for example).
+            # If we're not going to replace a previous notification
+            # we should withdraw said stale previous notification.
+            self._app.withdraw_notification(self._app_id)
         else:
-            notification.add_button(_('Play'), 'app.play')
-        notification.add_button(_('Skip'), 'app.next-song')
+            song = window.current_song
+            # This matches GNOME-Shell's format
+            notification = Gio.Notification.new(song.artist)
+            # GNOME focuses the application by default, we want to match that behavior elsewhere such as on KDE.
+            notification.set_default_action('app.activate')
+            notification.set_body(song.title)
 
-        if self._last_id != song.trackToken:
-            self._app.withdraw_notification(self._last_id)
+            if song.artUrl:
+                icon = Gio.FileIcon.new(Gio.File.new_for_uri(song.artUrl))
+            else:
+                icon = self._fallback_icon
+            notification.set_icon(icon)
 
-        self._last_id = song.trackToken
-        self._app.send_notification(song.trackToken, notification)
+            notification.add_button(_('Skip'), 'app.next-song')
+
+            self._app.send_notification(self._app_id, notification)
 
     def on_disable(self):
-        if self._last_id:
-            self._app.withdraw_notification(self._last_id)
-            self._last_id = ''
-
-        self.window.disconnect(self.song_change_handler)
-        self.window.disconnect(self.state_change_handler)
-
+        self._app.withdraw_notification(self._app_id)
+        if self._song_change_handler:
+            self.window.disconnect(self._song_change_handler)
+            self._song_change_handler = 0
+        if self._shutdown_handler:
+            self._app.disconnect(self._shutdown_handler)
+            self._shutdown_handler = 0


JasonLG1979:
Works On Xubuntu 18.10 but I get a couple unrelated warnings:
```
(pithos:10872): Gtk-WARNING **: 08:36:11.509: Failed to register client: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Method "RegisterClient" with signature "ss" on interface "org.xfce.Session.Manager" doesn't exist


** (pithos:10872): WARNING **: 08:36:12.225: Binding 'XF86AudioPlay' failed!
ERROR - mediakeys:on_enable:149 - Failed to bind media keys with Keybinder
```

TingPing:
> (pithos:10872): Gtk-WARNING **: 08:36:11.509: Failed to register client: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Method "RegisterClient" with signature "ss" on interface "org.xfce.Session.Manager" doesn't exist

Looks like that is in the next XFCE release (4.14):
https://github.com/xfce-mirror/xfce4-session/commit/894894991b90c3deab7a3ba230aafb6c91ca42f6#diff-20c3923f824ffafdc166a0bd58dbb57a

JasonLG1979:
I'm starting to question the value of the play / pause button in the notifications. Pretty much all of the DE's come with builtin MPRIS integration and all of the *buntus do. Really is a user likely to see a notification for a new song pop up and say "I want to pause this song"? More than likely in most cases they're going to dismiss the notification, raise Pithos to do whatever or skip the song. Pause and Play can be done at anytime via the main UI, mediaKeys or MPRIS. It would also allow for less notifications and for us to reuse the Gio.Notification object.

TingPing:
I think I agree that pause is an odd function to expose there but I imagine some people are accustomed to it. Do any other players do the same?

pithos/plugins/notify.py
```diff
@@ -50,25 +48,20 @@ def send_notification(self, window, *ignore):
         notification = Gio.Notification.new(song.artist)
         notification.set_body(song.title)
         if song.artUrl:
-            notification.set_icon(Gio.FileIcon.new(Gio.File.new_for_uri(song.artUrl)))
+            icon = Gio.FileIcon.new(Gio.File.new_for_uri(song.artUrl))
+        else:
+            icon = Gio.ThemedIcon.new_with_default_fallbacks('audio-x-generic-symbolic')
```
TingPing: The backend doesn't support fallbacks, it hardcodes the first value: https://github.com/GNOME/glib/blob/5a73eef7498bacce9b5b176a927af8ad20d2947e/gio/gfdonotificationbackend.c#L302-L304

JasonLG1979:
> I think I agree that pause is an odd function to expose there but I imagine some people are accustomed to it. Do any other players do the same?

Most players don't expose any functionality in notifications.

pithos/plugins/notify.py
```diff
@@ -50,25 +48,20 @@ def send_notification(self, window, *ignore):
...
         else:
             notification.add_button(_('Play'), 'app.play')
         notification.add_button(_('Skip'), 'app.next-song')
 
-        if self._last_id != song.trackToken:
-            self._app.withdraw_notification(self._last_id)
-
-        self._last_id = song.trackToken
-        self._app.send_notification(song.trackToken, notification)
+        self._app.send_notification('Pithos Song Change Notification', notification)
```
TingPing: The id isn't a user exposed value, this can be simplified to just our app-id or something that we know is unique.

TingPing:
Then 👍 to removing Play/Pause at least.

pithos/plugins/notify.py
JasonLG1979: Ok then I'll just use ```Gio.ThemedIcon.new('audio-x-generic')```

GNOME Shell auto-upgrades to symbolic icons.

pithos/plugins/notify.py
JasonLG1979: OK app id it is then

pithos/plugins/notify.py
TingPing: > GNOME Shell auto-upgrades to symbolic icons.

Oh nice.

pithos/plugins/notify.py
```diff
@@ -25,7 +24,9 @@ class NotifyPlugin(PithosPlugin):
     description = 'Shows notifications on song change'
 
     _app = None
-    _last_id = ''
+    _app_id = None
+    _notification = None
+    _fallback_icon = None
```
TingPing: __[Flake8]__ <sub>posted by Sider</sub>

[W291] trailing whitespace

JasonLG1979:
Something like that then? I'll test it out on all the above DE's.

TingPing:
I question the value of keeping `self._notification` around.

JasonLG1979:
> I question the value of keeping `self._notification` around.

If we can reuse the object why not?

JasonLG1979:
Just a guess but Gio.Notification appears to just be a property container.

TingPing:
Yes it just saves us allocating a new object each time. So if we agree that isn't meaningful the question is just which looks like more elegant code.

JasonLG1979:
> Yes it just saves us allocating a new object each time. So if we agree that isn't meaningful the question is just which looks like more elegant code.

You know that I am not the master or authority when it comes to elegant code,lol!!! It's your call.

TingPing:
I think it is more logically clear to build a new Notification each time you send one even if not technically necessary. You can keep around the fallback icon I suppose since logically it never changes.

JasonLG1979:
The other reason for keeping the notification object around is so we don't have to set the same button over and over again since it never changes either.

JasonLG1979:
So more like this then?

TingPing:
Seems good.

JasonLG1979:
Amended to account for stale notifications if we aren't going to replace an existing notification.

pithos/plugins/notify.py
```diff
@@ -35,40 +35,36 @@ def on_prepare(self):
...
-        if song.artUrl:
-            notification.set_icon(Gio.FileIcon.new(Gio.File.new_for_uri(song.artUrl)))
-
-        if window.playing:
-            notification.add_button(_('Pause'), 'app.pause')
+            # GNOME-Shell will auto dismiss notifications
+            # when the window becomes "active" but other DE's may not (KDE for example).
+            # If we're not going to replace a previous notification
+            # we should withdraw said stale previous notification.
+            self._app.withdraw_notification(self._app_id)
```
TingPing: This seems like an odd place to do this. If you want the notification to go away with window focus then you should connect to window focus?

JasonLG1979: It's not to make notifications go away when the widow focuses it's to dismiss notifications from previous songs if we're not going to replace them. GNOME Shell just happens to auto dismiss notifications when the app is "active".

JasonLG1979: We probably also want to withdraw any notifications on shutdown.

JasonLG1979: Something like:

```python
    def do_shutdown(self):
        self.withdraw_notification(self.get_application_id())
        Gtk.Application.do_shutdown(self)
        if self.window:
            self.window.destroy()
```

TingPing: Yea, the plugin can connect to the shutdown signal.

JasonLG1979: Again GNOME will auto dismiss notifications when we shutdown (with the freedesktop backend) KDE does not.

JasonLG1979:
@TingPing, Another nice little KDE "difference" is that unless we explicitly set the default action clicking on our notifications don't raise Pithos...

JasonLG1979:
So far the solution that I've come up with for that is to create a rather redundant "activate" app action and set it as the default action.

```python
        action = Gio.SimpleAction.new("activate", None)
        action.connect("activate", lambda action, param: self.do_activate())
        self.add_action(action)

....

    notification.set_default_action('app.activate')
```

It has no adverse effect on GNOME.

TingPing:
Looks good. I wonder if we should change that behavior in GLib itself.

TingPing:
Hmm, probably not as some apps may want it to do literally nothing.

JasonLG1979:
> Hmm, probably not as some apps may want it to do literally nothing.

So far it only does nothing on KDE. On all other DE's so far Pithos is raised as expected without the need to set the default action.

JasonLG1979:
@TingPing, Another observation. On many DE's notifications cover the screen area where the DE's MPRIS widget is shown and the "click-though-ability" of notifications varies by DE. So you end up with situations like in Xubuntu where you can't trigger the default action of the notification while it's MPRIS widget is shown it just clicks though. Or basically the opposite on Pantheon where you can't interact with the MPRIS widget unless you dismiss or activate the notification. Basically I'd like to make it so that if the song is changed via MPRIS no notification is sent since the user presumably has a MPRIS widget open thus making the notification redundant.

TingPing:
> Basically I'd like to make it so that if the song is changed via MPRIS no notification is sent since the user presumably has a MPRIS widget open thus making the notification redundant.

This becomes a problem is DEs start doing keybindings that toggle MPRIS.

JasonLG1979:
> This becomes a problem is DEs start doing keybindings that toggle MPRIS.

True. I guess there's only so much we can do to headlock DE's into having consistent behaviour,lol!!!

JasonLG1979:
OK. Only Cinnamon let to test and this is all set.

JasonLG1979:
It's all good. Notifications work more or less as expected on all the DE's I tested on.

JasonLG1979:
The MPRIS Implementation on Linux Mint Cinnamon show's Pithos' name as "Io" though. If I had to guess they're using the DesktopEntry prop for the display name and not the Identity prop. I'll file a bug report.

TingPing:
>  If I had to guess they're using the DesktopEntry prop for the display name and not the Identity prop. 

Well, that would be fine if they used that prop correctly and you know.. parsed the desktop file. Splitting it at first `.` is just completely dumb.

TingPing:
Oh actually they probably are using the bus name, which is more understandable but still very wrong.

JasonLG1979:
> Oh actually they probably are using the bus name, which is more understandable but still very wrong.

They do. I filed a bug.

https://github.com/linuxmint/Cinnamon/issues/8061

JasonLG1979:
@TingPing, Well like I said this works as expected unless you have anything else that needs changing it's ready to merge :)

JasonLG1979:
Merge me....

pithos/application.py
```diff
@@ -87,12 +87,9 @@ def do_startup(self):
...
 
-        action = Gio.SimpleAction.new("play", None)
-        action.connect("activate", lambda action, param: self.window.user_play())
-        self.add_action(action)
-
-        action = Gio.SimpleAction.new("pause", None)
-        action.connect("activate", lambda action, param: self.window.user_pause())
+        # Needed for notifications to function as expected in KDE.
+        action = Gio.SimpleAction.new("activate", None)
+        action.connect("activate", lambda action, param: self.do_activate())
```
TingPing:
```suggestion
        action.connect("activate", lambda action, param: self.activate())
```

pithos/application.py
```diff
@@ -192,6 +189,7 @@ def do_activate(self):
         self.window.present()
 
     def do_shutdown(self):
+        self.withdraw_notification(self.get_application_id())
```
TingPing: Move this to the plugin.

pithos/plugins/notify.py
```diff
@@ -35,40 +35,38 @@ def on_prepare(self):
...
         else:
-            notification.add_button(_('Play'), 'app.play')
-        notification.add_button(_('Skip'), 'app.next-song')
+            song = window.current_song
+            # This matches GNOME-Shell's format
+            notification = Gio.Notification.new(song.artist)
+            # Needed for KDE otherwise clicking on our notifications does nothing...
```
TingPing:
```suggestion
            # GNOME focuses the application by default, we want to match that behavior elsewhere such as on KDE.
```

pithos/plugins/notify.py
```diff
@@ -35,40 +35,38 @@ def on_prepare(self):
...
-            self._app.withdraw_notification(self._last_id)
-            self._last_id = ''
-
-        self.window.disconnect(self.song_change_handler)
-        self.window.disconnect(self.state_change_handler)
+            self._app.send_notification(self._app_id, notification)
 
+    def on_disable(self):
+        self._app.withdraw_notification(self._app_id)
+        self.window.disconnect(self._song_change_handler)
```
TingPing: `self._song_change_handler` may be `0`.

pithos/plugins/notify.py
JasonLG1979: No problem.

pithos/plugins/notify.py
JasonLG1979: Sure I'll do a ```if``` to check for that but it never should be 0. ```connect``` always returns a non-zero number and ```on_disable``` is always preceded by ```on_enable```. There is nothing in ```on_prepare``` or  ```on_enable``` that should ever fail.

pithos/application.py
JasonLG1979: How? by connecting to the shutdown signal?

The docs for the shutdown signal say:

"The ::shutdown signal is emitted only on the registered primary instance immediately after the main loop terminates"

I'm not sure how we're suppose to withdraw an notification with no mainloop?

pithos/application.py
JasonLG1979: No problem

pithos/plugins/notify.py
TingPing: `connect` can return `0` on failure but yes it will probably never fail in our usage. Just call it defensive programming =)

pithos/application.py
TingPing: That is odd phrasing that indeed sounds wrong.

JasonLG1979:
Well @TingPing I think that covers all the changes you asked for.

PR #579, opened 2018-11-25
Add PULSE_PROP_application.id for the sake of completeness

JasonLG1979:


JasonLG1979:
@TingPing, Does Flatpak do anything currently to confine multimedia? I've heard of Pipewire seems like an interesting and ambitious project. From what I understand it's suppose to be a replacement for pulse and jack on the audio side and be a way to sandbox multimedia?

PR #581, opened 2018-11-27
Fix MPRIS plugin hide on close pref

JasonLG1979:
Disabling the option and then enabling it didn't work and would result in quiting Pithos on close.

This makes sure to disconnect from on_destroy when we enable after a disable.

### DIFF
--- a/pithos/plugins/mpris.py
+++ b/pithos/plugins/mpris.py
@@ -837,7 +837,7 @@ class MprisPluginPrefsDialog(Gtk.Dialog):
 
         self.pithos = window
         self.settings = settings
-        self.window_delete_handler = None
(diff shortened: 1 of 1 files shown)


JasonLG1979:
@TingPing Not sure what's so "Complex" about the code? (the CodeFactor fail)

Anyway the notification icon plugin more than likely suffers the same issue with enabling after a disable. And as a side note the 2 plugins conflict because they can either end up breaking each other by connecting to on_destroy when disabled. I'm currently looking for a way for either to know if the other is suppose to hide on close so they don't step on each other.

JasonLG1979:
Or you know we could just get rid of the notification icon plugin,lol...

PR #593, opened 2019-09-18
Put album art in a directory that is cleaned up even if Pithos is killed

alexhenrie:
Today I found that over the course of 1 year and 9 months I have accumulated 213 pithos temp directories in ~/.cache, totalling 365 MiB. There doesn't seem to be any advantage to putting album art in ~/.cache instead of /tmp, and there is a significant disadvantage because files in ~/.cache are not automatically deleted if Pithos is not properly closed.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -366,8 +366,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.worker = GObjectWorker()
 
         try:
-            self.tempdir = tempfile.TemporaryDirectory(prefix='pithos-',
-                                                       dir=GLib.get_user_cache_dir())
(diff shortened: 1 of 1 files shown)


TingPing:
Of course there is a benefit to using cache, it persists to save downloading it the future, and it would save memory usage.

We just need to try automatically cleaning it up I think.

alexhenrie:
@TingPing As it stands, ~/.cache/pithos-* is never checked for previously downloaded album art. All album art is redownloaded every time Pithos runs.

TingPing:
Hmm, you're right we shouldn't be making a random dir and should always use `.cache/pithos` and clean up the contents.

PR #606, opened 2020-03-23
Add quit to the app menu.

JasonLG1979:


JasonLG1979:
@TingPing the reasoning behind this is that it's possible for users in DE's besides GNOME to not have an obvious way to quit Pithos if the hide instead of quit mpris plugin pref is selected.

JasonLG1979:
I've been meaning to add this for a while, but it really doesn't affect me since I use GNOME,lol!!!

data/ui/PithosWindow.ui
```diff
@@ -30,6 +30,13 @@
         <attribute name="label" translatable="yes">_About</attribute>
       </item>
     </section>
+    <section>
+      <item>
+        <attribute name="action">app.quit</attribute>
+        <attribute name="label" translatable="yes">_Quit</attribute>
+            <attribute name="accel">&lt;Primary&gt;q</attribute>
```
TingPing:
```suggestion
        <attribute name="accel">&lt;Primary&gt;q</attribute>
```

TingPing:
This broke the keybinding for me.

TingPing:
Moving them all to code works, so I guess I'll just do that.

Issue #595, opened 2019-10-15
Media Keys Will Not Bind in KDE (even with Keybinder)


trey-jones:
What baffles me is that Play/Pause/Skip worked fine on a clean install of Manjaro KDE, even without Keybinder.  At some point either through an update or installing some other software, possibly enabling MPRIS/KDEConnect (which works fine) it stopped working.  Here are the relevant logs:

```INFO - plugin:enable:53 - Preparing module mediakeys
DEBUG - mediakeys:on_new_finish:112 - DBus binding failed
INFO - plugin:_enable:59 - Enabling module mediakeys

** (pithos:518): WARNING **: 08:53:34.056: Binding 'XF86AudioPlay' failed!
ERROR - mediakeys:on_enable:149 - Failed to bind media keys with Keybinder
```

Pacman info (from AUR):

```Name            : pithos
Version         : 1.5.0-1
Description     : Native Pandora Radio client
Architecture    : any
URL             : https://pithos.github.io/
Licenses        : GPL3
Groups          : None
Provides        : None
Depends On      : gtk3  python-gobject  libsecret  python-cairo  gst-plugins-good  gst-libav  gst-plugins-base
Optional Deps   : libkeybinder3: for media keys plugin [installed]
                  libappindicator-gtk3: Unity indicator applet support [installed]
                  python-pacparser: PAC proxy support
                  python-pylast: Last.fm scrobbling support
                  python-systemd: Logging to the system journal
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : None
Installed Size  : 780.00 KiB
Packager        : Unknown Packager
Build Date      : Mon 14 Oct 2019 05:59:10 PM EDT
Install Date    : Mon 14 Oct 2019 05:59:12 PM EDT
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : None
```
Same happens on master - I've done some light debugging, and I might have time to make a PR if someone shoves me in the right direction.  I'm fairly new to KDE and I'm not sure if it's as simple as adding something to `de_busnames` or whether it's more of a KDE problem than a Pithos problem.

JasonLG1979:
X or Wayland? Because I'm not sure that Keybinder works with Wayland? Otherwise we don't do anything to bind the mediaKeys in KDE. I'm not even sure they have a DBus interface specifically for binding mediaKeys. Last I looked they used MPRIS for mediaKey binding.

TingPing:
[...]

PR #608, opened 2020-03-26
Changed name of method from SupportedUriScheme to SupportedUriSchemes to

obsidian74:
Changed name of method from SupportedUriScheme to SupportedUriSchemes to
conform to the spec at:
https://specifications.freedesktop.org/mpris-spec/latest/Media_Player.html#Property:SupportedUriSchemes

This fixes integration with KDE's Media Player framework

Addresses Issue: Media Keys Will Not Bind in KDE (even with Keybinder) #595


JasonLG1979:
Well that would explain that. Good catch!!!

Issue #611, opened 2020-03-31
User action is required to resume playback after a network failure


BHSPitMonkey:
If Pithos encounters a DNS resolution error (or other networking issues while changing songs or buffering) during its lifetime, it presents a modal dialog over its window that the user must acknowledge before resuming playback (even after connectivity has been restored). This also blocks attempts to resume or skip over MPRIS, Media Controls, keyboard shortcuts, etc. while in this state.

Short connectivity or DNS blips are common in my environments, so I find myself needing to stop what I'm doing and navigate back to the Pithos window to deal with this pretty often.

(Thanks for maintaining this wonderful project! :heart:)

TingPing:
I agree this isn't ideal.

Design wise it should probably just be an [info bar](https://developer.gnome.org/hig/stable/info-bars.html.en) letting you know, while it retries occasionally. Assuming we have enough width to fit one.

PR #612, opened 2020-04-01
WIP: Show non-modal infobar for some errors

TingPing:
TODO:
- [ ] Since this is no longer modal we need to hide the infobar when the error is bypassed. I guess we effectively need to track a "connected" state.
- [ ] We may want to retry automatically in some situations
- [ ] We can show better buttons, like on auth failure show preferences not retry
- [ ] Some pandora errors deserve the dialog?
- [ ] Ellipsize long messages, and add tooltip for full message?

The pandora error messages frankly suck, maybe we should add our own, bit out of scope for this PR...

See also #611

### DIFF
--- a/data/ui/PithosWindow.ui
+++ b/data/ui/PithosWindow.ui
@@ -234,6 +234,39 @@
           </object>
         </child>
         <child>
+          <object class="GtkInfoBar" id="error_infobar">
+            <property name="visible">1</property>
+            <property name="message-type">warning</property>
+            <property name="revealed">0</property>
+            <child internal-child="content_area">
+              <object class="GtkBox">
+                <property name="visible">1</property>
+                <child>
+                  <object class="GtkLabel">
+                    <property name="visible">1</property>
+                  </object>
+                </child>
+              </object>
+            </child>
+            <child internal-child="action_area">
+              <object class="GtkBox">
+                <property name="visible">1</property>
+                <child>
+                  <object class="GtkButton" id="info_button_retry">
+                    <property name="label" translatable="yes">Retry</property>
+                    <property name="visible">1</property>
+                    <property name="can_default">1</property>
+                    <property name="receives_default">1</property>
+                  </object>
+                </child>
+              </object>
+            </child>
+            <action-widgets>
+              <action-widget response="0">info_button_retry</action-widget>
+            </action-widgets>
+          </object>
+        </child>
+        <child>
           <object class="GtkScrolledWindow">
             <property name="visible">1</property>
             <property name="vadjustment">adjustment</property>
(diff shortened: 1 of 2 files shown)


JasonLG1979:
> The pandora error messages frankly suck, maybe we should add our own, bit out of scope for this PR...

In the event of a connection issue those errors more than likely aren't pandora errors, after all it means we can't connect to pandora. Those are urllib.errors. We could maybe translate them into less technical terms?

JasonLG1979:
> * Since this is no longer modal we need to hide the infobar when the error is bypassed. I guess we effectively need to track a "connected" state.

We already track if we're connected. `pandora.connected`. In the event of some pandora api errors and pretty much all urllib.errors we can basically assume that we're no longer connected.

> * We may want to retry automatically in some situations

We can do that pretty easily with `pandora_reconnect`

> * We can show better buttons, like on auth failure show preferences not retry

Again pretty easy. Pretty much all of the known pandora api error codes are already enumerated.

> * Some pandora errors deserve the dialog?

In the event of network errors we should show the error in the infobar and retry a couple times with long enough times between retries that the network has a little time to recover. Maybe like 3 tries a min apart then show a dialog?

Pandora api errors are a whole different animal then network errors. There are a few that should not result in a retry and some that no amount of reties will help. Again though most of the common api errors already have sub messages in the `ApiError` class.

JasonLG1979:
Off the top of my head:

These errors should not be retied and should pretty much be considered fatal at least temporarily:
`MAINTENANCE_MODE`
`INSUFFICIENT_CONNECTIVITY`
`READ_ONLY_MODE`
`PARTNER_NOT_AUTHORIZED`
`DEVICE_MODEL_INVALID`

These errors should present a "hey dummy" dialog as these are the direct result of user interaction: 
`MAX_STATIONS_REACHED` *non fatal you just can't add anymore stations.
`PLAYLIST_EXCEEDED` *fatal temporarily.

These errors should present the appropriate user setting and retry after user interaction:
 `COUNTRY_NOT_SUPPORTED` *proxy settings
`INVALID_LOGIN` *prefs dialog.

TingPing:
If you want to see this work through that would be great, I'm not sure I'll find the time.

JasonLG1979:
> If you want to see this work through that would be great, I'm not sure I'll find the time.

Sure, I'll see what I can come up with next weekend. With the THING going on I'm trying to get my kids setup for at home learning this weekend. They haven't said so yet but I think school is done for the year. I'm also finalizing a thorough (read: tweaker) cleaning routine for them since they basically can't leave the house and I'll be damned if I have to ride out the apocalypse in a dirty house,lol!!!

PR #614, opened 2020-04-23
sync loved songs with last.fm

hoffoo:
this PR is to send loved tracks to last.fm 

i didn't add the corresponding code to "unlove" because banning a song from a station doesn't necessarily mean that you don't like.

let me know if you think this is a good addition


TingPing:
Hmm, I vaguely remember this being a thing many years ago and there were complaints that the two just don't mean the same thing for a lot of people.

This probably just needs to be put behind a setting and for now we can default it off.

hoffoo:
old PR

TheKrafter:
@hoffoo, can this not just be a config option in the plugin preferences that defaults to off (as TingPing suggested)?

From what TingPing said it seems this could be merged after doing so.

PR #628, opened 2020-08-31
Adding in an activate handler for search box - resolves #597

### LINKED ISSUE #597: <Return> to select top result of search
Alternatively, perhaps <Return> could only do something if there is only one station in the results.

This might be related to #561 (although I haven't checked if I can reproduce that one) in the sense that it has to do with <Return>.

tylerthetiger:
The purpose of this PR is to allow users to accept the first search result when filtering stations with the Enter key instead of having to click on the result.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -416,7 +416,7 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.stations_popover.set_model(self.stations_model)
         self.stations_popover.listbox.connect('row-activated', self.active_station_changed)
         self.stations_button.set_popover(self.stations_popover)
-        self.stations_popover.search.connect('activate',self.search_activate_handler)
+        self.stations_popover.search.connect('activate', self.search_activate_handler)
 
     def init_actions(self, app):
         action = Gio.SimpleAction.new('playpause', None)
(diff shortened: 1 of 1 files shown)


pithos/pithos.py
```diff
@@ -1250,6 +1251,14 @@ def destroy_ui_loop(self):
             GLib.source_remove(self.ui_loop_timer_id)
             self.ui_loop_timer_id = 0
 
+    def search_activate_handler(self,station):
+        row = self.stations_popover.listbox.get_row_at_y(0)
+        if not row:
+            return False
+        self.active_station_changed(self.stations_popover.listbox,row)
```
TingPing, line 1258:
```suggestion
        self.station_changed(row.station)
```

pithos/pithos.py
```diff
@@ -416,6 +416,7 @@ def init_ui(self):
         self.stations_popover.set_model(self.stations_model)
         self.stations_popover.listbox.connect('row-activated', self.active_station_changed)
         self.stations_button.set_popover(self.stations_popover)
+        self.stations_popover.search.connect('activate',self.search_activate_handler)
```
TingPing, line 419:
```suggestion
        self.stations_popover.search.connect('activate', self.search_activate_handler)
```

pithos/pithos.py
```diff
@@ -1250,6 +1251,14 @@ def destroy_ui_loop(self):
             GLib.source_remove(self.ui_loop_timer_id)
             self.ui_loop_timer_id = 0
 
+    def search_activate_handler(self,station):
+        row = self.stations_popover.listbox.get_row_at_y(0)
+        if not row:
+            return False
```
TingPing, line 1257:
```suggestion
            return
```

pithos/pithos.py
```diff
@@ -1250,6 +1251,14 @@ def destroy_ui_loop(self):
...
+        self.active_station_changed(self.stations_popover.listbox,row)
+        self.stations_popover.hide()
+        self.stations_popover.search.set_text('')
```
TingPing, line 1260: Maybe the hide logic should be moved to `StationsPopover.py` as it is for the row activation. No strong opinion.

tylerthetiger: Updated it to just call `on_row_activated`.

PR #633, opened 2020-10-31
ui: Make fit on small (mobile) screens

BrainBlasted:
Adjust width requests and widgets to fit on mobile screens. 
Removes the toolbar in favor of a GtkBox, as the toolbar
prevented proper sizing and is removed in GTK4.


TingPing:
Thanks!

PR #634, opened 2020-11-01
CI: add a flatpak job

bilelmoussaoui:
The job is running on my fork at https://github.com/bilelmoussaoui/pithos/runs/1339100384

### DIFF
--- a/flatpak/io.github.Pithos.json
+++ b/flatpak/io.github.Pithos.json
@@ -1,7 +1,7 @@
 {
   "app-id": "io.github.Pithos",
   "runtime": "org.gnome.Platform",
-  "runtime-version": "master",
+  "runtime-version": "3.38",
   "sdk": "org.gnome.Sdk",
   "command": "pithos",
   "finish-args": [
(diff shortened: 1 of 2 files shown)


flatpak/io.github.Pithos.json
```diff
@@ -1,7 +1,7 @@
 {
   "app-id": "io.github.Pithos",
   "runtime": "org.gnome.Platform",
-  "runtime-version": "master",
+  "runtime-version": "3.38",
```
BrainBlasted, line 4: Why 3.38?

bilelmoussaoui: Why not? the GNOME docker image could be used though which should work fine I guess. But as Pithos is not really targeting the latest GNOME, it makes sense to use the latest stable version.

TingPing: I agree we don't really target master, this seems reasonable.

TingPing:
Just throwing a thought out there:

I would love to remove the LGTM and SideCI integrations for a generic GitHub action that runs a python linter.

Issue #555, opened 2017-12-23
HeaderBar


ForLoveOfCats:
In the course of the ongoing redesigning of Pithos the decision was made to more closely follow the Gnome Shell HIG. While in many ways this is a good thing, one feature of the Gnome Shell HIG, the HeaderBar, will cause Pithos to look out of place on any Desktop Environment other that Gnome.

In an effort to ensure that Pithos does not lose favor with non-Gnome users I took it upon myself to find a way for the HeaderBar to be used without the unfortunate side effects. I believe I have done so.

I have written a wrapper around HeaderBar, with an almost one to one API, that will allow for HeaderBar to be enabled/disabled with one function call and no duplicated code. I have released this wrapper under GPL 2 and request that the Pithos community look into using either it or the ideas it is built upon.

Wrapperbar: https://github.com/ForLoveOfCats/wrapperbar

JasonLG1979:
window.get_position() doesn't currently work with wayland.

JasonLG1979:
I also get this:
```
Gtk-WARNING **: gtk_window_set_titlebar() called on a realized window
```

TingPing:
I would make a Gtk.ApplicationWindow subclass that automatically uses the expected one so it can just be used in the gtkbuilder file.

That said I'm not convinced we really need to support the old layout as it will just look bad using a toolbar or something at the top.

ForLoveOfCats:
I should have documented that Gtk-WARNING better in the project. That simply means that Gtk must reinitialize the window. All window contents are preserved so there are no side effects. Unfortunately there is no way around this.

ForLoveOfCats:
```That said I'm not convinced we really need to support the old layout as it will just look bad using a toolbar or something at the top.```

Using a toolbar looks arguably better than a HeaderBar on non-Gnome platforms especially as DE's such as XFCE and Plasma 5 use their own systems for handling Window Decorations which this would allow Pithos to respect while still using HeaderBar if the user wishes.

ForLoveOfCats:
I am aware much code in this library is more than likely naive or not up the the quality standards of Pithos. If you find anything wrong, low quality, or a missing feature please feel free to send me a PR.

JasonLG1979:
I really don't see the benefit of moving the contents of the header bar below the title bar. One of the main complaints of the header bar is that it makes the title bar "fat" and takes up more space. This doesn't solve that at all.

[...]

PR #635, opened 2020-11-01
PithosWindow: Use GtkHeaderBar

BrainBlasted:
Makes Pithos use a GtkHeaderBar with client side
decorations. Moves the song controls and menu
into the HeaderBar.

Closes https://github.com/pithos/pithos/issues/555

### DIFF
--- a/data/ui/PithosWindow.ui
+++ b/data/ui/PithosWindow.ui
@@ -48,171 +48,179 @@
     <property name="icon_name">io.github.Pithos</property>
     <signal name="configure-event" handler="on_configure_event" swapped="no"/>
     <signal name="destroy" handler="on_destroy" swapped="no"/>
-    <child>
-      <object class="GtkBox">
+    <child type="titlebar">
+      <object class="GtkHeaderBar">
         <property name="visible">1</property>
-        <property name="orientation">vertical</property>
+        <property name="can_focus">False</property>
+        <property name="show_close_button">True</property>
         <child>
-          <object class="GtkBox">
+          <object class="GtkBox" id="playcontrol_box">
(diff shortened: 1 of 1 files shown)


TingPing:
One thing lost is the current song in the titlebar. I know it doesn't really fit in the headerbar so I agree with removing it however the listview doesn't really do a great job of highlighting the current playing track.

This isn't necessarily a blocker but I don't know how much UI work you are intending to do. Ideally we would be moving to GtkListBox and maybe that concern could be handled then.

BrainBlasted:
I'm willing to do a decent amount of UI work - perhaps a play icon could be shown on the current song?

TingPing:
I'd have to look at some mock ups, but yeah a simple solution like that makes sense. I was also just thinking of making an actual highlight color to the current playing distinct from current selection but maybe that is a bit odd.

BrainBlasted:
Colors are a bit hard to do "properly" with GTK, so an additional color would be a challenge, imo.

TingPing:
A simple icon indicating it sounds good. I'll merge this for now.

Thanks for these contributions!

JasonLG1979:
I like this!!!

@TingPing maybe increase the default starting width a tiny bit? There's a lot going on in the headerbar, it would be nice to give the station names a little more room?

This is a default size of 550 x 430. The extra height makes 4 tracks fit neatly in Adwaita and Yaru.

pithos-resize

pithos-resize2

TingPing:
@JasonLG1979 Sounds good to me.

JasonLG1979:
@TingPing I put in a PR for it.

PR #651, opened 2021-04-28
Update flatpak manifest json and shared-modules submodule

rcalixte:
Add libnotify as notifications do not consistently work without it and update shared-modules submodule


TingPing:
Pithos doesn't use libnotify at all.

PR #680, opened 2022-08-08
Fix notifications for Flatpak and add album info

### LINKED ISSUE #677: Song Change Notifications do not work in Flatpak version
Tested on 2 different 20.04 Pop_OS Machines using the Flatpak version, the song change notifications when enabled in the plugins do not seem to work.

I tested with and without enabling the "Notification Icon" plugin and making sure its closed/minimized to the system tray before the song change.

rcalixte:
In the Flatpak app, Gio.Notification.set_icon() does not seem to work with Gio.FileIcon types where Gio.ThemedIcon does.
Adding album info to the notification as well (perhaps this could be an option in the preferences if this is undesired).
Using a more Pythonic ternary operator

Resolves #677


pithos/plugins/notify.py
```diff
@@ -52,16 +52,15 @@ def send_notification(self, window, *ignore):
...
             song = window.current_song
+            album = '\n' + song.album if song.album else ''
             # This matches GNOME-Shell's format
             notification = Gio.Notification.new(song.artist)
             # GNOME focuses the application by default, we want to match that behavior elsewhere such as on KDE.
             notification.set_default_action('app.activate')
-            notification.set_body(song.title)
+            notification.set_body(song.title + album)
+
+            icon = Gio.ThemedIcon.new(song.artUrl) if song.artUrl else self._fallback_icon
```
TingPing, line 62: I'm confused how this works tbh, sounds like a bug.

We need to download the data and use `Gio.BytesIcon` actually.

rcalixte: The `artUrl` property is already a local file URI in the cache. Why would we want to re-download it?

rcalixte: @TingPing I've tried with `Gio.BytesIcon` as well but to no avail, so possibly a bug upstream. Can we merge until the upstream bug is resolved?

TingPing: No. As-is this honestly makes 0 sense.

Themed Icons are icons defined here: https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

Using an HTTP url here is nonsense and shouldn't work or be used.

rcalixte: The `artUrl` property is just a file:// URI to the local filesystem. It is downloaded into the cache as the playlist populates. Why doesn't it make sense to reuse the same locally cached files? I'm a little confused on this one.

rcalixte: @TingPing L39 in this file... `_fallback_icon` has been a `ThemedIcon` for quite some time now.

rcalixte: @TingPing I agree in the unorthodox nature but this is the only `Gio.Icon` implementation I could get to work successfully with `Gio.Notification.set_icon()`.

rcalixte: Would you prefer to remove the album art in favor of just the fallback icon? This would solve the original issue of notifications not working on the Flatpak version of the app just the same.

TingPing: I would prefer just using BytesIcon. I'm sure it will work.

rcalixte: Do you want to look through some of the force-pushed changes that include the `Gio.BytesIcon` attempts? Maybe it's tangential to the same bug or maybe not, but it is not working. I'll play with it more later but it does not feel like it will trend toward the outcome of success.

TingPing: Hmm. You're usage of BytesIcon seems correct.

TingPing: Testing this locally I can confirm `BytesIcon` didn't work. However your `ThemedIcon` changes also do not work here.

GNOME-Shell 42.3 - xdg-desktop-portal 1.12.6

TingPing: So `xdg-desktop-portal` accepts only BytesIcon or ThemedIcon (indeed FileIcon as before couldn't work) - https://github.com/flatpak/xdg-desktop-portal/blob/main/src/notification.c#L281

TingPing: `glib` does nothing except serialize it as-is: https://github.com/GNOME/glib/blob/main/gio/gportalnotificationbackend.c#L67

rcalixte:
```bash
$ apt list gnome-shell xdg-desktop-portal
Listing... Done
gnome-shell/jammy-updates,now 42.2-0ubuntu0.2 amd64 [installed,automatic]
xdg-desktop-portal/jammy-updates,now 1.14.4-1ubuntu2~22.04.1 amd64 [installed,automatic]
```

I tested that the icon does not show on GNOME but that the notifications do work. (GNOME also has a redundant MPRIS interface so it may not be as much of an issue there.) If this introduces a regression in notifications on GNOME, I can try to workaround that as well.

On other systems, I have Cinnamon 5.4 and Enlightenment 0.24.2 running and can confirm that the notifications do work from the Flatpak installation where they did not previously.

TingPing: I would guess this is a GNOME-Shell bug but tbh I don't see anything off: https://github.com/GNOME/gnome-shell/blob/main/js/ui/notificationDaemon.js

rcalixte: This looks interesting: https://github.com/GNOME/gnome-shell/blob/main/js/ui/notificationDaemon.js#L70

It could be a bug in Gio.

TingPing: This codepath is not hit for us.

`xdg-desktop-portal-gtk` calls `org.gtk.Notifications` handled here: https://github.com/GNOME/gnome-shell/blob/main/js/ui/notificationDaemon.js#L462

What you linked is when `org.freedesktop.Notifications` is called.

TingPing: Oh just saw your other comment. On Cinnamon/Enlightenment maybe they just don't handle BytesIcons.

rcalixte: Cinnamon is forked from GNOME 3 so I don't expect too much variance there. Enlightenment, I would need to follow up.

rcalixte: When I tested on GNOME 42, the notifications worked but they presented without the icons. Is that the main blocker to merging for now?

TingPing: Cinnamon is a very different codebase. They for example do not listen on `org.gtk.Notifications` at all so its behavior is 100% different: https://github.com/linuxmint/cinnamon/blob/master/js/ui/notificationDaemon.js

Anyway yes I'd want it to at least work on GNOME-Shell.

rcalixte: > Cinnamon is a very different codebase. They for example do not listen on `org.gtk.Notifications` at all so its behavior is 100% different: https://github.com/linuxmint/cinnamon/blob/master/js/ui/notificationDaemon.js

Fascinating since it is working fine for me without modifying the talk-name configuration.


> Anyway yes I'd want it to at least work on GNOME-Shell.

The notifications worked for me when I tested on GNOME 42 (sans album photo). Can you share some details with what issues you saw on your end? Is there anything in the logs?

TingPing: > Fascinating since it is working fine for me without modifying the talk-name configuration.

pithos -> xdg-desktop-portal -> xdg-desktop-portal-gtk -> if available org.gtk.Notifications, otherwise org.freedesktop.Notifications. No permission required.

> The notifications worked for me when I tested on GNOME 42 (sans album photo)

Sorry if unclear. The missing album art is the only issue.

rcalixte: > > The notifications worked for me when I tested on GNOME 42 (sans album photo)
> 
> Sorry if unclear. The missing album art is the only issue.

Can you confirm whether notifications were working on the current stable Flatpak build as well? I believe notifications from Flatpak were entirely broken (after the initial notification on startup if the window was hidden before playback). If notifications were indeed working on GNOME, we can test for that and use the existing code. Also, if it is just the Flatpak version and GNOME, we can test for that as well. I can add comments around the conditionals and tests we introduce until this is cleaned up upstream. I'll also test with both native and Flatpak builds with GNOME on my local installation when I have the time.

rcalixte: @TingPing I tested with a local build and Flatpak and it looks like only the Flatpak for GNOME needs changes as the compiled version worked as expected. Please shout if you have observed anything different from this as I'm getting started on separating out that condition now and hope to have an updated PR for review soon.

rcalixte: @TingPing Please feel free to test against the latest PR. This should be good to go for all environments. I added a comment to the code for future contributors as well.

pithos/plugins/notify.py
```diff
@@ -34,6 +34,9 @@ def on_prepare(self):
         if not os.path.exists('/.flatpak-info'):
             os.environ['GNOTIFICATION_BACKEND'] = 'freedesktop'
 
+        _env = os.environ['XDG_SESSION_DESKTOP'].lower()
+        self._gnome_flatpak_env = True if os.path.exists('/.flatpak-info') and 'gnome' in _env else False
```
TingPing, line 38: See `utils.py` for `is_flatpak()`

pithos/plugins/notify.py
```diff
@@ -52,16 +55,24 @@ def send_notification(self, window, *ignore):
...
             song = window.current_song
+            album = '\n' + song.album if song.album else ''
             # This matches GNOME-Shell's format
             notification = Gio.Notification.new(song.artist)
             # GNOME focuses the application by default, we want to match that behavior elsewhere such as on KDE.
             notification.set_default_action('app.activate')
-            notification.set_body(song.title)
+            notification.set_body(song.title + album)
 
+            # Use BytesIcon for GNOME and Flatpak and ThemedIcon everywhere else, otherwise notifications do not work
```
TingPing, line 65: I still really hate this. Its very wrong that this works this way.

Let's do:

```python
notify_server_name = ... 

if is_flatpak():
  # Use Gio.BytesIcon. Anything else from a sandbox is nonsense.
else:
  notify_server_name = ...# Get the result of `org.freedesktop.Notifications.GetServerInformation()` (This should be async, so that adds some complications)
  if notify_server_name in ('whatever', ): # A list of daemons where Gio.FileIcon does not work?
    # FIXME: This is a weird hack where we shove a file URI in a themed icon and some daemons accept this and use it as they do not support FileIcon.
    # Use Gio.ThemedIcon
  else:
    # Use Gio.FileIcon
```

TingPing: Also I'm sorry if I keep repeating myself. But at the end of the day it is *not* Pithos' bug that `Gio.BytesIcon` doesn't work somewhere. That is a real bug outside of Pithos and should be fixed there.

`Gio.ThemedIcon` working at all is truly a hack. One I can accept as long as its reported upstream where the real bug is, limited in scope only to where it is known to work, and documented for what it is.

rcalixte: I'm not 100% sure where the bug should be filed upstream or what details would need to be provided. I understand and agree that a longer-term solution is to address the upstream bug rather than work around it but it feels outside of the scope of addressing the problem in the present. I can open an issue where needed and provide what I can but it is unlikely I would be able to own it which gives me further hesitation. (I'm not a GNOME user primarily after all.)

And the scope of working and not working is inverted to my understanding. `BytesIcon` doesn't "work" in GNOME, it just doesn't prevent the notification from being displayed. On Cinnamon and Enlightenment (unsure of the scope outside of these two), it causes notifications not to be shown **at all** when Pithos is installed via Flatpak, negatively impacting the end user experience as evidenced by the open Issues.

Breaking this further down into two issues:

1. Address the bug with notifications in the current application as it is affecting end users.
2. Implement the fix in line with the expectations of the available libraries.

The fix for 1 is to use `FileIcon` for the version of Pithos packaged outside of Flatpak. When running as a Flatpak, if GNOME is the desktop, use `BytesIcon` and in all other Flatpak conditions, use `ThemedIcon` for now, as the scope is everything not GNOME that is impacted. (I also realize this was not written into the conditionals well in the latest commit but I am working on the requested changes now as well as correcting the logic to match.) This fixes the issue immediately for impacted end users.

When the libraries are patched upstream and known to be working as designed, we can address 2 and should revisit the code and re-implement it as it is intended to function. That change would and should be both transparent to the end user and free from regression.

Can you agree on this strategy and path forward?

rcalixte: @TingPing Can you share your thoughts on the latest push and the previous comments above? The alternative approach you've suggested is the inverse strategy to solve this problem. It might be better to test or solicit feedback from more desktop environments before concluding that GNOME is the norm rather than the exception, maybe via a pinned Issue for users to submit feedback. The idea of a tuple (or list or array or set) that would need to be maintained until some upstream bug is resolved is a significantly more tiresome and errorful approach. Between KDE Plasma, MATE, XFCE, LXDE, and many others, this alternative approach potentially leaves them in a broken state which is not the preferred strategy.

If it would be easier and your schedule is amenable to it, we can set a time to discuss this over something like Jitsi. Or we can continue with this thread as is. The main motives being to ensure that we are on the same page and that we can reach an unbiased conclusion in a timely fashion.

pithos/plugins/notify.py
```diff
@@ -52,20 +57,30 @@ def send_notification(self, window, *ignore):
...
             song = window.current_song
+            album = '\n' + song.album if song.album else ''
             # This matches GNOME-Shell's format
             notification = Gio.Notification.new(song.artist)
             # GNOME focuses the application by default, we want to match that behavior elsewhere such as on KDE.
             notification.set_default_action('app.activate')
-            notification.set_body(song.title)
+            notification.set_body(song.title + album)
 
+            # FIXME: Use BytesIcon for Flatpak GNOME and ThemedIcon as a workaround other DEs and Flatpak,
```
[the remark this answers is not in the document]
TingPing: Testing this again neither works on GNOME. I'm very confused.

pithos/plugins/notify.py
TingPing: At this point I'll accept whatever works in the most places. It is a frustrating problem.

pithos/plugins/notify.py
```diff
@@ -52,20 +57,30 @@ def send_notification(self, window, *ignore):
             self._app.withdraw_notification(self._app_id)
         else:
             song = window.current_song
+            album = '\n' + song.album if song.album else ''
             # This matches GNOME-Shell's format
             notification = Gio.Notification.new(song.artist)
             # GNOME focuses the application by default, we want to match that behavior elsewhere such as on KDE.
             notification.set_default_action('app.activate')
-            notification.set_body(song.title)
+            notification.set_body(song.title + album)
```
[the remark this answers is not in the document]
rcalixte: Can do. I can open a smaller and separate PR with these changes.

pithos/plugins/notify.py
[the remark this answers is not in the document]
rcalixte: Did something change with your setup? I can test again in maybe a few days on GNOME. On the whole, it's a little bit disheartening as I thought we were closer to the finish line.

TingPing:
I made a simple test and at first it didn't work, but restarting `xdg-desktop-portal` and now it just works (weird...):

```python
#!/usr/bin/env python3

import sys
from gi.repository import Gio

print('Showing a notification for icon: {}'.format(sys.argv[1]))

def on_activate(app):
    notification = Gio.Notification.new('Testing')

    f = Gio.File.new_for_path(sys.argv[1])
    data, _ = f.load_bytes(None)
    icon = Gio.BytesIcon.new(data)
    notification.set_icon(icon)
    app.send_notification(None, notification)
    app.hold()

# Pithos has a desktop file with `X-GNOME-UsesNotifications`.
app = Gio.Application.new('io.github.Pithos', Gio.ApplicationFlags.FLAGS_NONE)
app.connect('activate', on_activate)
app.run()
```

TingPing:
The exact same code does not work in the middle of Pithos though...

rcalixte:
> The exact same code does not work in the middle of Pithos though...

Hmm, I haven't had a chance to circle back yet but this adds yet another wrinkle. I do want to get a fix merged so I'll try to set aside some time soon to revisit this.

rcalixte:
@TingPing Just an update here, I've updated my systems to flatpak 1.14.0 and xdg-desktop-portal 1.14.4 and while there is still no album art with any of the preferred implementations (BytesIcon or FileIcon), I can confirm that notifications are working with what I think is the fallback icon. This is built from a locally rebased dev branch. I'll update with more as I have it. Thanks again!

PR #681, opened 2022-08-30
Add album info to notifications

rcalixte:
Adding album info to notifications


rcalixte:
@TingPing Thank you for the patience. This is now ready for review.

pithos/plugins/notify.py
```diff
@@ -77,3 +80,47 @@ def on_disable(self):
...
+    def __init__(self, window, settings):
+        super().__init__(use_header_bar=1)
+        self.set_title(_('Notification Preferences'))
+        self.set_default_size(300, -1)
+        self.set_resizable(False)
+        self.connect('delete-event', self.on_close)
+
+        self.pithos = window
+        self.settings = settings
+        self.show_album = literal_eval(self.settings['data']) if self.settings['data'] else False
```
TingPing, line 97: Let's not do `literal_eval`. A simple `bool()` to get it and `str()` to store it is plenty.

rcalixte: The issue is that Boolean strings don't convert to bool types easily in Python.

```python
>>> bool('True')
True
>>> bool('False')
True
```

rcalixte: I can try switching it to string comparison.

pithos/plugins/notify.py
```diff
@@ -77,3 +80,47 @@ def on_disable(self):
...
+
+        box = Gtk.Box()
+        label = Gtk.Label()
+        label.set_markup('<b>{}</b>\n{}'.format(_('Show Album Info'), None))
```
TingPing, line 101:
```suggestion
        label.set_markup('<b>{}</b>\n'.format(_('Show Album Info')))
```

Surely you meant this.

pithos/plugins/notify.py
```diff
@@ -77,3 +79,47 @@ def on_disable(self):
...
+        self.set_title(_('Notification Preferences'))
+        self.set_default_size(300, -1)
+        self.set_resizable(False)
+        self.connect('delete-event', self.on_close)
+
+        self.pithos = window
+        self.settings = settings
+        self.show_album = self.settings['data'] == 'True' if self.settings['data'] else False
+
+        box = Gtk.Box()
```
TingPing, line 98: It looks very squished together, you should set the margin. Like `6` all around maybe?

pithos/plugins/notify.py
```diff
@@ -77,3 +79,47 @@ def on_disable(self):
...
+
+        self.switch = Gtk.Switch()
+        self.switch.connect('notify::active', self.toggle_album_info)
+        self.switch.set_active(self.show_album)
+        self.settings.connect('changed::enabled', self._on_album_info_plugin_enabled)
+        self.switch.set_halign(Gtk.Align.END)
+        self.switch.set_valign(Gtk.Align.CENTER)
+        box.pack_end(self.switch, False, False, 2)
+
+        content_area = self.get_content_area()
```
TingPing, line 112: This is deprecated, can you just do `self.add(box)`? I forget if that works if you could test.

pithos/plugins/notify.py
TingPing: Ah. Yes a simple string comparison is good.

pithos/plugins/notify.py
rcalixte: ```(pithos:2): Gtk-WARNING **: 13:00:25.803: Attempting to add a widget with type GtkBox to a NotifyPluginPrefsDialog, but as a GtkBin subclass a NotifyPluginPrefsDialog can only contain one widget at a time; it already contains a widget of type GtkBox```

I'll test with it though.

pithos/plugins/notify.py
rcalixte: It was weirdly generating errors previously because it needed a second parameter. That seems to have resolved so I've removed the second parameter and the new-line while adding a valign setting to the label as well.

pithos/plugins/notify.py
rcalixte: The spacing helped but I just added a pair of new-lines to the label. I think it helps how it looks?

pithos/plugins/notify.py
rcalixte: The variable naming is different but the latest (assuming it's up to date since it was last edited over two years ago) [example](https://python-gtk-3-tutorial.readthedocs.io/en/latest/dialogs.html#example) suggests the structure is correct.

rcalixte: I see the deprecation notice with GTK4. I've switched it from a GtkDialog to a GtkWindow object and that seems to be working fine minus one issue where the preference is displayed away from the plugins and main application windows. I'll keep looking to see if I can resolve this.

rcalixte: I think this is just a byproduct of using Gtk.Window instead of Gtk.Dialog. Window managers will position the Window object the same as other window objects. The other preferences are still using Gtk.Dialog so this will create a bit of inconsistency in terms of user experience.

I'll commit whichever you prefer. If it's Gtk.Window, this is ready to merge as is. If it is Gtk.Dialog, I can push the previous state.

rcalixte: I thought of another thing. If Gtk.Window is the direction you want things to go while also wanting the user experience to be uniform, I can work on PRs for all of the other plugins with preferences. Then the PRs could be review/merged simultaneously.

rcalixte:
@TingPing Will you have a chance to provide feedback on direction here? I believe this is ready to merge but I can make any edits if needed as well.

Issue #686, opened 2022-11-04
Pithos cache grows without bound


zaporter:
## Problem
I'm a very long time pithos fan and I was just making a backup when I realized that: 
`du -hs ~/.var/app/io.github.Pithos/cache/tmp` returns 47Gib
`find -iname "*.jpeg" | wc -l` returns 275067
## Expected
Pithos should not grow the cache forever and should only cache a few Gib at most. Having more than a quarter million jpegs feels wrong. 

## Solution
I could just run `rm -rf` every few months with a crontab, however this behavior still feels like a bug. Would it be possible to add some kind of cache clearing algorithm? If you don't want to write this, I could do it if you tell me how you want it implemented.

P.S. I am a huge fan of yours, TingPing. Thank you for your work.

TingPing:
@jarebear6expepjozn6rakjq5iczi3irqwphcvb Hey a PR for this would be good.

There is actually a second problem. The cache files are never actually read as a cache. The directory they go into changes each run.

I agree it should be a core feature. For now I wouldn't even include settings for it.

go140point6:
This fix breaks running Pithos on Debian (Bullseye) tested with a Pi 4 (64bit OS) and Pi Zero W (32bit OS) when compiling from source.  I've never been very successful with Flatpak on the Pi, so get Pithos running from source.  Perhaps the version of the tempfile module needs to be updated in Bullseye, but to get Pithos v1.6.1 running I removed the "ignore_cleanup_errors=True" flag from pithos/pithos.py line 372:

self.tempdir = tempfile.TemporaryDirectory(prefix='pithos-', dir=tempdir_base, ignore_cleanup_errors=True)

That flag seems to be unknown to whatever version Bullseye is running.

PR #692, opened 2023-03-02
fix: delete temp files on exit 

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
I added the `ignore_cleanup_errors=True` flag to make sure the `tempfile` library stayed honest and actually deleted the directories it created.  This prevents the buildup of directories and files and closes issue #686.

If you want to actually use this as a cache I would remove the `tempfile` library and opt for a simpler system for file governance.  You explicitly set `tempdir_base` without it anyway, and there seems to be no other reliance on it elsewhere.

Is actual caching a feature you/others want? Any other meta-data that should be cached?

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -1076,7 +1076,6 @@ class PithosWindow(Gtk.ApplicationWindow):
         self.current_song.duration = self.query_duration() or self.current_song.trackLength * Gst.SECOND
         self.current_song.duration_message = self.format_time(self.current_song.duration)
         self.update_song_row()
-        self.check_if_song_is_ad()
         # We can't seek so duration in MPRIS is just for display purposes if it's not off by more than a sec it's OK.
         if self.current_song.get_duration_sec() != self.current_song.trackLength:
             self.emit('metadata-changed', self.current_song)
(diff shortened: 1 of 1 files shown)


jarebear6expepjozn6rakjq5iczi3irqwphcvb:
Sorry, not Gtk gripes but rather because the `tempfile.NamedTemporaryFile(prefix='art-', suffix='.jpeg', dir=tmpdir.name, delete=False) ` object has a false delete flag cause the errors preventing the cleanup.   Not sure if you had a reason set that flag, but either way this PR fixes it.

TingPing:
Could you update the commit message to make sense.

I think re-using the downloaded files would be good but that can be a later change.

TingPing:
You never actually updated the commit message...

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
How is that?

TingPing:
The issue title is not the commit message. `git commit -am 'This is the new message'` will do it. then `git push -f`

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
Got it I think. Thank you

PR #699, opened 2023-06-22
feat: add caching support to album art files

### LINKED ISSUE #686: Pithos cache grows without bound

### LINKED ISSUE #698: TemporaryDirectory API used is too new for Debian (Bullseye)
> This fix breaks running Pithos on Debian (Bullseye) tested with a Pi 4 (64bit OS) and Pi Zero W (32bit OS) when compiling from source.  I've never been very successful with Flatpak on the Pi, so get Pithos running from source.  Perhaps the version of the tempfile module needs to be updated in Bullseye, but to get Pithos v1.6.1 running I removed the "ignore_cleanup_errors=True" flag from pithos/pithos.py line 372:

```
self.tempdir = tempfile.TemporaryDirectory(prefix='pithos-', dir=tempdir_base, ignore_cleanup_errors=True)
```

> That flag seems to be unknown to whatever version Bullseye is running.

_Originally posted by @go140point6 in https://github.com/pithos/pithos/issues/686#issuecomment-1601819379_

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
My take on adding cache support for album art removes reliance on the tempfile library and creates a static directory for art to be managed in. 

Currently clears files older than 15 days (1.296e+6 seconds).  Was not sure when to run this check so it clears cache every time artwork is requested. Please advise if there is a more sane spot for it to trigger.

fixes #686 
fixes #698

This also fixes cases where pithos is closed in a way other than the gui (x) which results in tempdir created directories to not delete themselves and multiply.

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -14,6 +14,7 @@
 
 
 import contextlib
+import hashlib
 import html
 import json
 import logging
@@ -60,6 +61,8 @@ except AttributeError:
 
 ALBUM_ART_SIZE = 96
 TEXT_X_PADDING = 12
+# Time in seconds to retain album art files
+ART_CACHE_TIME = 1.296e+6
 
 FALLBACK_BLACK = Gdk.RGBA(red=0.0, green=0.0, blue=0.0, alpha=1.0)
 FALLBACK_WHITE = Gdk.RGBA(red=1.0, green=1.0, blue=1.0, alpha=1.0)
@@ -369,9 +372,8 @@ class PithosWindow(Gtk.ApplicationWindow):
                 # However in flatpak that path is not readable by the host.
                 tempdir_base = os.path.join(GLib.get_user_cache_dir(), 'tmp')
             self.tempdir = os.path.join(tempdir_base, 'pithos_cache')
-            if not os.path.exists(self.tempdir):
-                os.makedirs(self.tempdir)
-                logging.info("Created temporary directory {}".format(self.tempdir.name))
+            os.makedirs(self.tempdir, exist_ok=True)
+            logging.info("Created temporary directory {}".format(self.tempdir))
         except IOError as e:
             self.tempdir = None
             logging.warning('Failed to create a temporary directory: {}'.format(e))
@@ -851,11 +853,12 @@ class PithosWindow(Gtk.ApplicationWindow):
 
     def clear_art_cache(self):
         logging.info('Checking for expired art in cache')
+        timestamp = time.time()
         with os.scandir(self.tempdir) as art_list:
             for art in art_list:
-                age = time.time() - art.stat().st_mtime
-                if age > 1.296e+6:
-                    os.remove(os.path.join(self.tempdir, art.name))
+                age = timestamp - art.stat().st_mtime
+                if age > ART_CACHE_TIME:
+                    os.remove(os.path.join(self.tempdir, art.path))
 
     def get_playlist(self, start = False):
         if self.playlist_update_timer_id:
@@ -889,15 +892,13 @@ class PithosWindow(Gtk.ApplicationWindow):
             song, index = extra
             if tmpdir:
                 try:
-                    # TODO move cache clear call somewhere more sane?
                     self.clear_art_cache()
-                    encoded_filename = (song.artist+song.album).encode('utf-8').hex()[:222]+'.jpeg'
-                    if os.path.exists(os.path.join(self.tempdir, encoded_filename)):
-                        file_url = urllib.parse.urljoin('file://', urllib.parse.quote(encoded_filename))
+                    filename_hash = hashlib.sha256((song.artist+song.album).encode('utf-8')).hexdigest()+'.jpeg'
+                    if os.path.exists(os.path.join(self.tempdir, filename_hash)):
+                        file_url = urllib.parse.urljoin('file://', urllib.parse.quote(filename_hash))
                     else:
-                        with open(os.path.join(self.tempdir, encoded_filename), 'xb') as f:
+                        with open(os.path.join(self.tempdir, filename_hash), 'xb') as f:
                             f.write(image)
-                            file_url = urllib.parse.urljoin('file://', urllib.parse.quote(f.name))
                 except IOError:
                     logging.warning("Failed to write art tempfile")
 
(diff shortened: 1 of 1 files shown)


pithos/pithos.py
```diff
@@ -369,8 +368,10 @@ def init_core(self):
             if is_flatpak():
                 # However in flatpak that path is not readable by the host.
                 tempdir_base = os.path.join(GLib.get_user_cache_dir(), 'tmp')
-            self.tempdir = tempfile.TemporaryDirectory(prefix='pithos-', dir=tempdir_base, ignore_cleanup_errors=True)
-            logging.info("Created temporary directory {}".format(self.tempdir.name))
+            self.tempdir = os.path.join(tempdir_base, 'pithos_cache')
+            if not os.path.exists(self.tempdir):
+                os.makedirs(self.tempdir)
```
TingPing, line 373: This is a race condition, just always call `os.makedirs()` with `exist_ok=True`.

pithos/pithos.py
```diff
@@ -848,6 +849,14 @@ def playpause_notify(self, *ignore):
         else:
             self.user_play()
 
+    def clear_art_cache(self):
+        logging.info('Checking for expired art in cache')
+        with os.scandir(self.tempdir) as art_list:
+            for art in art_list:
+                age = time.time() - art.stat().st_mtime
+                if age > 1.296e+6:
```
TingPing, line 857: Could you move this number into a documented global variable.

pithos/pithos.py
```diff
@@ -848,6 +849,14 @@ def playpause_notify(self, *ignore):
         else:
             self.user_play()
 
+    def clear_art_cache(self):
+        logging.info('Checking for expired art in cache')
+        with os.scandir(self.tempdir) as art_list:
+            for art in art_list:
+                age = time.time() - art.stat().st_mtime
```
TingPing, line 856: You can cache `time.time()` instead of repeatedly calling it.

pithos/pithos.py
```diff
@@ -848,6 +849,14 @@ def playpause_notify(self, *ignore):
...
+                if age > 1.296e+6:
+                    os.remove(os.path.join(self.tempdir, art.name))
```
TingPing, line 858: `art.path` should work.

pithos/pithos.py
```diff
@@ -877,11 +886,18 @@ def get_album_art(url, tmpdir, *extra):
...
             file_url = None
+            song, index = extra
             if tmpdir:
                 try:
-                    with tempfile.NamedTemporaryFile(prefix='art-', suffix='.jpeg', dir=tmpdir.name, delete=False) as f:
-                        f.write(image)
-                        file_url = urllib.parse.urljoin('file://', urllib.parse.quote(f.name))
+                    # TODO move cache clear call somewhere more sane?
+                    self.clear_art_cache()
+                    encoded_filename = (song.artist+song.album).encode('utf-8').hex()[:222]+'.jpeg'
```
TingPing, line 894: I'm assuming the `:222` is to limit the length? Just use a hash like sha256 instead?

pithos/pithos.py
```diff
@@ -877,11 +886,18 @@ def get_album_art(url, tmpdir, *extra):
...
+                    if os.path.exists(os.path.join(self.tempdir, encoded_filename)):
+                        file_url = urllib.parse.urljoin('file://', urllib.parse.quote(encoded_filename))
+                    else:
+                        with open(os.path.join(self.tempdir, encoded_filename), 'xb') as f:
+                            f.write(image)
+                            file_url = urllib.parse.urljoin('file://', urllib.parse.quote(f.name))
```
TingPing, line 900: `file_url` is the same in both conditions right? Could be lifted out?

TingPing (changes requested):
Thanks for doing this, its a huge improvement! Just a few comments that could be changed.

pithos/pithos.py
```diff
@@ -889,15 +892,13 @@ def get_album_art(url, tmpdir, *extra):
...
                 try:
-                    # TODO move cache clear call somewhere more sane?
                     self.clear_art_cache()
-                    encoded_filename = (song.artist+song.album).encode('utf-8').hex()[:222]+'.jpeg'
-                    if os.path.exists(os.path.join(self.tempdir, encoded_filename)):
-                        file_url = urllib.parse.urljoin('file://', urllib.parse.quote(encoded_filename))
+                    filename_hash = hashlib.sha256((song.artist+song.album).encode('utf-8')).hexdigest()+'.jpeg'
+                    if os.path.exists(os.path.join(self.tempdir, filename_hash)):
+                        file_url = urllib.parse.urljoin('file://', urllib.parse.quote(filename_hash))
                     else:
```
TingPing:
```suggestion
                    file_url = urllib.parse.urljoin('file://', urllib.parse.quote(filename_hash))
                    if not os.path.exists(os.path.join(self.tempdir, filename_hash)):
```

Shouldn't this be like this?

`file_url` is always the same, just write the file if not cached already?

pithos/pithos.py
```diff
@@ -60,6 +61,8 @@
 
 ALBUM_ART_SIZE = 96
 TEXT_X_PADDING = 12
+# Time in seconds to retain album art files
```
TingPing, line 64:
```suggestion
# 15 days in seconds to retain album art files.
```

pithos/pithos.py
TingPing: Oh actually isn't this path just wrong. It is `os.path.join(self.tempdir, filename_hash)` not just `filename_hash`.

jarebear6expepjozn6rakjq5iczi3irqwphcvb: Yeah I think I got it this time

Issue #679, opened 2022-06-21
GStreamer 1.20 regression causes Pithos to stop playing after 1 or 2 songs


dooberdoober:
I have allowed pithos to play enough to make sure its an issue. Computer was updated and restarted yesterday. Now, pithos either wont play at all when I open it and start a station. Or it will play through one song, maybe two, and then stop. The time on the song it should be playing next just sits at zero and there have been a few songs that played but the time stayed at 0:00 during the entire track.

I have no other evidence to provide at this time; I was hoping there was already an issue opened for this but I seem to be the only one. 
Im running current version of Manjaro KDE with 5.10.123-1 kernel
I do not have a paid pandora sub but I have successfully/happily used Pithos for at least 3 years.

Any ideas or assistance are appreciated!
Certainly so if this is just me and no one else.

nms-scribe:
I'm having the same problem on Arch (kernel 5.18.1-arch1-1) using XFCE. Pianobar seems to be working.

user4815162342:
I'm having the same problem on Arch (kernel 5.18.1-arch1-1, XFCE). Pithos 1.5. I discovered it on 6/13 after running an update. Pianobar works though.

kalekundert:
I've been having the same problem too, recently (pithos 1.5.1, Arch 5.18.7-arch1-1, openbox).  The one thing I have to add is that pithos will start playing again if you skip the song it's stuck on.

dooberdoober:
Correct.
I was witnessing the issue for a few weeks on and off (it seemed) but I was by the computer or a tethered device and could easily click next. Its when I started to spend days with it playing and able to hear but not close so I would have to walk over. I dealt with it for awhile as it wasn't that big of a deal.....but then I really started to wonder if it was just me and it was starting to reliably stop play after each track and thats when I sat and tested it before opening this issue.

sartwell:
I am having the same problem with 22.04 Kubuntu.  I thought it was the power saving on the cheap laptop i am running it on, but I noticed other people are having the same issue.

FractalMachinist:
This has been happening to me over the last week or two. I could skip songs or change stations to resume playback, but that's stopped working as of today.

TingPing:
This is most likely a change in gstreamer, considering Pithos hasn't changed in those years.

Somebody will need to spend a good amount of time looking into it.

user4815162342:
Is there any way to put a temporary workaround in Pithos? 

[...]

PR #704, opened 2024-02-27
Avoid racecondition when buffering that appears to halt Pithos

fenugeek:
Displaying the remaining song time in the GUI is triggered from a per-second timer. If the position in the stream is queried exactly at the time a buffering refresh is finishing, Pithos sometimes ends up permanently pointing somewhere that doesn't answer queries or alert to an end-of-song. On a congested network, this means Pithos appears to stop playing every song or two.

This fix is the barest minimum change, that avoids making the query to determine the stream position when we are in a buffering situation.

Fixes [#679](https://github.com/pithos/pithos/issues/679)

### DIFF
--- a/pithos/pithos.py
+++ b/pithos/pithos.py
@@ -1204,7 +1204,15 @@ class PithosWindow(Gtk.ApplicationWindow):
         album = html.escape(song.album)
         msg = []
         if song is self.current_song:
-            song.position = self.query_position()
+            # Avoid race condition where query_position() sometimes points us
+            # permanently to an invalid segment if we are called whilst
+            # buffering. Most reliable semaphore for buffering state is
+            # buffering_timer_id, cleared after buffering (see on_gst_buffering())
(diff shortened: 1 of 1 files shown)


pithos/pithos.py
```diff
@@ -1204,7 +1204,15 @@ def song_text(self, song):
         album = html.escape(song.album)
         msg = []
         if song is self.current_song:
-            song.position = self.query_position()
+            # Avoid race condition where query_position() sometimes points us
+            # permanently to an invalid segment if we are called whilst
+            # buffering. Most reliable semaphore for buffering state is
+            # buffering_timer_id, cleared after buffering (see on_gst_buffering())
+            if self.buffering_timer_id == 0:
+                song.position = self.query_position()
```
TingPing, line 1212: I'm not sure what an "invalid segment" is, but can you not change `query_position()` to never return that?

PR #727, opened 2026-05-29
Modernize: GTK3→GTK4 migration, GNOME 50 runtime, feature additions

DavidMcFarlin:
## Summary

This branch revives Pithos 1.6.3 on current desktops by porting it to
GTK4, refreshing the Flatpak runtime, and rounding out a handful of
Pandora features that were always in the JSON API but never exposed in
the UI.

### Modernization
- Retire the expired Pandora CA cert, harden networking, and bump
  Flatpak runtime + Python deps
- Phased GTK3 → GTK4 migration (templates, widget tree, gesture
  controllers, popover menus)
- Phase 3: custom `CellRendererAlbumArt` and icon-theme handling
- Runtime fixes plus `ComboBoxText` → `DropDown` migration
- Bump Flatpak runtime to GNOME 50 (47 went EOL 2025-10-15); vendor
  `typing-extensions` / `setuptools-scm` and disable pip build isolation
  so pylast still builds on Python 3.13
- Fix GTK4 media-key fallback

### Discoverability
- Always-visible hamburger menu
- One-click "New Station…" entry point
- "Why is This Playing?" via `track.explainTrack`

### Pandora feature additions
- Genre Station Browser (`station.getGenreStations`)
- Station Seed Manager — add / remove song, artist, genre seeds
  (`station.getStation`, `addMusic`, `deleteMusic`)
- Share Station via email (`station.shareStation`)
- Bookmarks viewer with delete (`user.getBookmarks` +
  `bookmark.delete*Bookmark`)
- Delete Station via right-click on the stations dropdown

### Polish
- Sensible default sizes on the new Phase B dialogs (they were opening
  at minimal height)
- Right-click context menu on the stations popover with a workaround
  for the GTK 4.22 nested-popover grab bug so the outer dropdown stays
  open while the context menu is shown

## Test plan

- [ ] `meson install` build + `pithos` on a host GTK 4.22 setup
- [ ] `flatpak-builder` build against GNOME 50 + `flatpak run`
- [ ] Login flow, station playback, skip/thumbs/tired
- [ ] Hamburger menu items wired correctly
- [ ] Genre Stations / Bookmarks / Edit Seeds / Share Station dialogs
      open at usable size and round-trip API calls
- [ ] Right-click a station → context menu visible alongside dropdown;
      Delete removes the row and falls forward if the deleted station
      was playing
- [ ] "Why is This Playing?" surfaces explanation traits
- [ ] Media keys still work (the GTK4 fallback path)

## Notes

Happy to split this into smaller PRs (modernization / discoverability /
feature additions / Flatpak runtime) if you'd prefer to review and land
them independently — the commits are already grouped along those lines.


TingPing:
> Modernize: retire expired CA cert, harden networking, bump deps and Flatpak runtime

This commit probably doesn't belong here and combines random things.

> [Features: genre browser, station seed manager, share station, bookmar…](https://github.com/pithos/pithos/pull/727/commits/b9f6d8b229a083e5d6a6e522a86e4acc8e7a334b)  
> [Features: discoverable hamburger menu, quick "New Station", "Why is t…](https://github.com/pithos/pithos/pull/727/commits/3d075514f989828ee9cffcb4163f8f6e63837dd7)

New features should be another PR.

> [flatpak: vendor typing-extensions and setuptools-scm for pylast build](https://github.com/pithos/pithos/pull/727/commits/0f0598ae1abfaf6670e986ff1fca37179160e0dd)
> [flatpak: bump runtime to GNOME 50](https://github.com/pithos/pithos/pull/727/commits/b9ec4d535e04994337a8e0610c574b9a3b560414)

Unless needed new PR.

> Ignore local dev-tool config and drop it from install set

This is harmless, but its also not needed at all, `bin/pithos` disables writing cache files. A `run` target exists to launch dev builds.

---

I understand that most of this is generated code but these need to have proper authors set.

I don't have much time to review things, but if it's limited to UI changes in one PR I likely can.

DavidMcFarlin:
Thanks for taking a look at the PR and for the feedback.

After reading your comments, I think the best path forward is for me to withdraw this PR rather than ask you to spend time reviewing and breaking it apart. My goal was primarily to get Pithos working reliably on current Linux desktops and Flatpak runtimes while exposing some Pandora functionality that was already available through the API. In hindsight, that ended up touching enough areas of the project that it no longer fits well into a single review.

I also understand that your available maintenance time is limited, and I don't want to create additional review or maintenance burden for a project that isn't currently under active development.

I'll continue maintaining these changes independently and take the project in a different direction rather than asking for further review here.

Thank you for maintaining Pithos over the years and for taking the time to look over the submission.

TheKrafter:
> I'll continue maintaining these changes independently and take the project in a different direction rather than asking for further review here.

Are you implying that you're no longer sending these contributions to Pithos, and making your own downstream version? It seems apparent to me that all these changes are very much welcome, but they should be in separate PRs rather than amalgamated here.

Would it not be better to split this into PRs like you offered to in your original comment, rather than splitting off into a separate project?

DavidMcFarlin:
> > I'll continue maintaining these changes independently and take the project in a different direction rather than asking for further review here.
> 
> Are you implying that you're no longer sending these contributions to Pithos, and making your own downstream version? It seems apparent to me that all these changes are very much welcome, but they should be in separate PRs rather than amalgamated here.
> 
> Would it not be better to split this into PRs like you offered to in your original comment, rather than splitting off into a separate project?

That's a fair question.

My concern isn't really about splitting the work into smaller PRs. The commits are already grouped and could be separated without too much difficulty.

What gives me pause is that the project appears to have very limited maintainer availability at the moment. The feedback I received suggests that most of the work would need to be split into multiple PRs and reviewed independently, which is completely reasonable from a project management perspective. At the same time, TingPing also mentioned having very little time available for review.

Given that combination, I'm not convinced it makes sense to create a long queue of dependent PRs and ask for additional review effort.

My original goal was simply to get Pithos working well on current Linux desktops and Flatpak runtimes. If there is active interest in reviewing and merging the modernization work in smaller pieces, I'm happy to reconsider. Otherwise, maintaining the changes independently may be the more practical path.

I certainly don't want this to come across as frustration with the review feedback. I appreciate the time everyone has taken to look at the changes.

TheKrafter:
I understand not wanting to overload TingPing with PRs, but I think it would be advantageous even if you do make a fork, for the end goal to be to merge with the original project, even if that means you have a forked version to distribute for awhile until TingPing can review each PR as they find the time to do so.

You've done a lot of good work here so far, and I think that keeping it within this project would be the most beneficial to its users and maintainers without undue fragmentation, as often happens in FOSS projects like this.


Ultimately, my main concern is that your fork leaves behind the expertise and experience here that maintainers like TingPing have, which has been very beneficial for the longevity of the project thus far.

Perhaps the concern instead should be looking for (an) additional maintainer(s)?

DavidMcFarlin:
I appreciate the perspective, and I don't disagree that keeping development within the existing project would generally be the best outcome.

To provide some context, my intention was never to create a competing project or fragment the user base. I'm a daily Pithos user myself. The reason I started making these changes was because I genuinely thought the project had become inactive.

From the outside, there hadn't been a release in roughly two years, issues and PRs appeared to be going unanswered, and there were multiple comments from users openly asking whether the project was dead. As someone who depends on the application, that was concerning. I wasn't seeing signs that anyone was actively working on it or planning future releases.

Because of that, I approached the work with the mindset of "what would I need to do to keep this application healthy and usable going forward?" rather than "what is small enough to get merged upstream?" The result was a broad set of changes that touched GTK4, Flatpak, build tooling, maintenance items, and Pandora functionality that I personally wanted to use.

The recent feedback is the first indication I've had that there is still interest in actively maintaining and reviewing changes. That's encouraging. At the same time, I don't think it's unreasonable for me to weigh that against the statement that review time is extremely limited.

My concern isn't that the work needs to be split up. It can be. My concern is investing significant effort into restructuring and defending a long series of PRs when the project's future direction and review bandwidth remain unclear.

If pithos is going to continue to be touted as an active project, it is going to need someone to step up and care about it. I'm not that person - I only did what I needed to do to keep the app functional, so I didn't have to worry about the next system update finally breaking it.

Issue #230, opened 2015-09-14
Windows / OS X Support and Installation Instructions?


giantspeck:
The changelog for Pithos mentions the addition of support for Windows (as of version 1.0.0, released on 1 May 2014) and OS X (as of version 1.0.1, released on 21 September 2014).

Has support for either of these operating systems ever been fully realized?  And if so, are there special instructions to install Pithos on either operating system?

TingPing:
> Has support for either of these operating systems ever been fully realized?

I would say no, the issues largely lie in GStreamer. Playback often fails or crashes on both Win32 and OSX.

A brief intro to installation though:

Windows:

- Use [msys2](https://msys2.github.io/) to install all deps or
- Use the [pygobjectwin32 project](http://sourceforge.net/projects/pygobjectwin32/)

OSX:

- Use [homebrew](http://brew.sh/) and [my package](https://github.com/TingPing/homebrew-gnome/blob/master/pithos.rb) (unmaintained) or
- Use [JHBuild](https://wiki.gnome.org/Projects/Jhbuild/Mac_OS) and [my patches](https://github.com/TingPing/jhbuild/tree/tingping/osx)

giantspeck:
Hey, just letting you guys know that [Hermes](https://github.com/HermesApp/Hermes) for OS X has pretty much abandoned their project and are willing to hand off development to anyone.

PR #729, opened 2026-09-15
Support macOS: store the account password in the Keychain

vlouvet:
This lets Pithos run on macOS from a Homebrew GTK 3 / GStreamer install. macOS has no Secret Service provider, so today Pithos stops at "You need to install a service such as gnome-keyring" (#230 has background on macOS support).

**Changes**
- `pithos/keychain.py` (new) stores the password in the login keychain through the system `security(1)` tool, so there are no new dependencies.
  - The command goes in on stdin (`security -i`), so the password never shows up in the process list. The password is passed as hex (`-X`), so it needs no quoting.
  - `security -i` exits 0 even when a command fails, so each write is checked by reading the item back.
- `pithos/util.py`: on `darwin`, `SecretService` is the Keychain backend and the `Secret` typelib isn't imported, so libsecret isn't needed there. The libsecret schema moved into `__init__` so the module still imports without it. I kept the platform switch simple, as you suggested on #697. Linux behaviour is unchanged.
- `pithos/plugins/notify.py`: don't force `GNOTIFICATION_BACKEND=freedesktop` on macOS, which has no fdo notification server.

The plugins that need DBus (MPRIS, media keys, journald logging) already handle the missing session bus: on macOS they fail on first run and switch themselves off.

This branch doesn't include #728, but macOS needs it too, because Homebrew's Adwaita 50 lacks `emblem-favorite-symbolic`.

**Tested**
- macOS 26, Apple Silicon (Homebrew GTK 3.24.52, GStreamer 1.28.7, Python 3.14):
  - Keychain round trip: store, update, rename the email, and a password with quotes, a backslash and non-ASCII characters.
  - Real Pandora login: stations load and a song starts streaming.
- Ubuntu 26.04, MATE with gnome-keyring 50:
  - The same round trip passes through the libsecret backend.
  - Startup matches master (MPRIS and media keys bind as before).
  - Real Pandora login works.
- `flake8` reports no new warnings.

Not tested: notifications on macOS (the plugin is off by default, and GLib may need an app bundle ID to show them), and `meson install` on macOS.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pithos/util.py
```diff
@@ -15,26 +15,27 @@
 
 import logging
 import os
+import sys
 from urllib.parse import splittype, splituser, splitpasswd
 
 import gi
-gi.require_version('Secret', '1')
+if sys.platform != 'darwin':
```
TingPing: Can move and reuse the `is_macos()`

pithos/keychain.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def _quote(value):
+    # Quoting rules of `security -i`'s command parser.
+    return '"' + value.replace('\\', '\\\\').replace('"', '\\"') + '"'
+
+
+def _lookup(email):
+    result = subprocess.run(
+        ['security', 'find-generic-password', '-a', email, '-s', _SERVICE, '-g'],
```
TingPing: I don't really love this design.

I'd rather you just pull in the `keyring` python package and let it do it. It uses the platforms [API](https://github.com/jaraco/keyring/blob/main/keyring/backends/macOS/api.py) instead of command line tools.

We could use the keyring library on windows and macos.

TingPing (changes requested):

pithos/util.py
vlouvet: Done. `is_macos()` now lives in `util.py` next to a new `is_windows()`, and `notify.py` imports it from there instead of testing `sys.platform` again.

pithos/keychain.py
vlouvet: Switched to `keyring`. `pithos/keychain.py` is gone; `pithos/keyring_backend.py` implements the same interface on top of it, so it goes through the platform API instead of `security(1)`, and Windows comes along via `WinVaultKeyring`.

Two things worth flagging:

- It is a new runtime dependency (MIT), but only on macOS and Windows — the import sits behind the platform check, so Linux packaging is untouched.
- Same service name and account as the previous version (`io.github.Pithos.Account` + the email), and I checked that keyring reads back items the `security(1)` code wrote, so anyone who tried the earlier branch keeps their password.

Tested with a round trip driving `KeyringService` directly — store, read, update in place, rename the account, look up a missing account, clean up — 13/13 on macOS (Security framework, plus a real Pandora login) and 13/13 on Windows 11 (Credential Manager). I have no GTK 3 stack on the Windows box, so only the credential path is proven there; I called that out in the description.

PR #730, opened 2026-09-15
CI: Switch to the maintained flatpak-builder action

vlouvet:
The Flatpak job is red on current PRs and on master's last run (a4e56f2), even though the build and the Meson tests pass. It fails only at the final step, uploading the bundle:

```
Error: Unable to create a container for the artifact pithos-devel at …/artifacts?api-version=6.0-preview
```

**CI** (4299fbe)
- `bilelmoussaoui/flatpak-github-actions@v2` still uploads through that retired artifact API. This switches to the maintained `flatpak/flatpak-github-actions/flatpak-builder@v6`, with the same inputs.
- v6 starts a session bus with `dbus-daemon`, which the old `bilelmoussaoui` image lacks, so the job now uses the Flathub image the action documents: `ghcr.io/flathub-infra/flatpak-github-actions:gnome-50`.
- The action now names the artifact `pithos-devel-x86_64.flatpak`.

**Runtime** (9835ba2, as requested)
- GNOME 41 → 50. GNOME 41 has been end-of-life since 2022, and 50 is the newest runtime on Flathub.
- pylast 2.2.0 + six → the pylast 4.5.0 wheel. The 2.2.0 sdist can't build offline on the SDK's Python 3.13, and the Last.fm plugin's pylast calls are unchanged.

**Tests** (cffe49b)
- The GNOME 50 SDK no longer ships `appstream-util`. Fall back to `appstreamcli` so "Validate appstream file" keeps running.

**Tested**
- The CI change alone, on my fork: https://github.com/vlouvet/pithos/actions/runs/35012853104
- The runtime bump, locally on Ubuntu 26.04 with flatpak-builder and the action's test setup:
  - the build works and all 10 tests pass
  - pylast reaches Last.fm inside the sandbox
  - a real Pandora login in the installed Flatpak works: the password is stored through the Secret portal, stations load and playback starts

With this merged, CI results on the open PRs (including #728 and #729) mean something again.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TingPing:
Can you update/test the flatpaks runtime also? https://github.com/pithos/pithos/blob/a4e56f299f4078964bd866c47bc8e6dfab537a99/flatpak/io.github.Pithos.json#L4

vlouvet:
Done: the runtime is now GNOME 50 (9835ba2). That's the newest on Flathub and the same as the CI image.

It needed one more change. On the GNOME 50 SDK (Python 3.13), the pinned pylast 2.2.0 sdist no longer installs offline, because pip's isolated build wants setuptools. I swapped `six` + pylast 2.2.0 for the pylast 4.5.0 wheel, which has no dependencies. Every pylast call the Last.fm plugin makes, including `scrobble`'s positional arguments, has the same signature in both versions, so the plugin is unchanged. Newer pylast releases pull in httpx, so I stopped at 4.5.0.

The GNOME 50 SDK also dropped `appstream-util`, which silently removed the "Validate appstream file" test. cffe49b falls back to `appstreamcli` so the test still runs, and the current appdata passes it.

Tested on Ubuntu 26.04 with flatpak-builder, using the same test setup as the action:
- the build works and all 10 tests pass
- inside the sandbox, pylast imports and reaches Last.fm (`auth.getToken`)
- a real Pandora login in the installed Flatpak works: the password is stored and read back through the Secret portal, stations load and playback starts

